PR #62, opened 2023-04-16
add change detection to limit plans/applies

verbanicm:
Adds the ability to detect git changes and only run guardian for those directories. 

A workflow job will be used to look for all terraform files with a `backend ".*" {` syntax. If a file exists in a directory with this syntax, it will be considered the root of a module and will be a target for plan/apply.

A second step will look for what files have changed and create the intersection between changed directories and the root directories.

### DIFF
--- a/.github/actions/directories/action.yml
+++ b/.github/actions/directories/action.yml
@@ -1,0 +1,74 @@
+# Copyright 2023 The Authors (see AUTHORS file)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: 'Guardian Directories'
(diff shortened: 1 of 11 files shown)


.github/actions/directories/action.yml
```diff
@@ -0,0 +1,81 @@
...
+    value: ${{ steps.get_dirs.outputs.dirs }}
+
+runs:
+  using: composite
+  steps:
+    - name: 'Get Directories'
+      id: 'get_dirs'
+      shell: 'bash'
+      run: |-
+        TERRAFORM_GLOB=$(echo "./${{ inputs.terraform_directory }}/**/*.tf" | sed 's|././|./|g' | sed 's|//|/|g')
```
bradegler, line 45: Are you sure you don't have to escape the `.`s in the `././`. I have a feeling that's not matching on what you think it is.

.github/actions/directories/action.yml
```diff
@@ -0,0 +1,81 @@
...
+        DIRS_WITH_BACKENDS=$(grep -E 'backend\s?".*"\s?{' $TERRAFORM_GLOB -l | xargs dirname | sed 's|^[\.?][/?]*||g' | sort | uniq)
```
bradegler, line 46: General request: can you put comments with your regexes to explain what they are doing? Mainly for future maintainers but it also aids in review.

verbanicm: ack

.github/actions/directories/action.yml
verbanicm: d'oh!

PR #68, opened 2023-04-18
move plan success logic to reusable workflow

verbanicm:


mattrandallbecker:
quoting those yaml strings just in case 👀

PR #84, opened 2023-05-22
adjust apply workflow ordering for errors

verbanicm:
Bootstrapping error file should come first, otherwise the following steps will fail.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -14,7 +14,7 @@
 
 locals {
   project_id = "guardian-i-42c69c"
-  name       = "test-change"
+  name       = "test-change-test"
 }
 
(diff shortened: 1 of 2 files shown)


terraform/main.tf
```diff
@@ -14,7 +14,7 @@
 
 locals {
   project_id = "guardian-i-42c69c"
-  name       = "test-change"
+  name       = "test-change-test"
```
sethvargo, line 17: Why this change?

sethvargo (approved):

terraform/main.tf
verbanicm: this is a "unit" test to force workflows to run the guardian repo.

PR #93, opened 2023-06-15
refactor(iam-drift): Remove third party action for issues.

dcreey:
Use actions/github-script instead.

Tested here: https://github.com/theteamatx/x-refinery-org/pull/158

### DIFF
--- a/.github/actions/iam-drift-detection/action.yml
+++ b/.github/actions/iam-drift-detection/action.yml
@@ -63,6 +63,12 @@ runs:
         echo "diff_detected=${diff_detected}" >> "$GITHUB_OUTPUT"
         echo "diff_file=${TMPFILE_DIFF}" >> $GITHUB_OUTPUT
 
+    - uses: 'actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce' # ratchet:actions/upload-artifact@v3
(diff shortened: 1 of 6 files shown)


.github/actions/iam-drift-detection/action.yml
```diff
@@ -47,34 +51,97 @@ runs:
...
+          await github.rest.issues.create({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            labels: [ISSUE_LABEL],
+            title: 'IAM drift detected',
+            body: issue_body,
+          });
 
+    - uses: 'actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410' # ratchet:actions/github-script@v6
+      id: 'maybe-close-previous-issue'
```
verbanicm, line 134: Why this and not updating the existing one?

dcreey: Arbitrarily picked based on what it was doing before. It does seem slightly easier to close and create a new one rather than having two create paths. But I don't feel strongly either way. If you want update lets do that in a follow up?

verbanicm: ACK, fine with this way.

verbanicm (approved):

PR #102, opened 2023-06-22
Do streaming json decoding

### LINKED ISSUE #97: [drift] Read stream directly into JSON unmarshaller for terraform state
### TL;DR

"It's really inefficient to download the entire object into a buffer, then unmarshal that buffer. It's better to stream the writer from the DownloadFileIntoMemory directly into a json.Decoder to save a lot of allocation and memory." - Seth

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

sethvargo:
Fixes #97

### DIFF
--- a/pkg/drift/gcs/gcs.go
+++ b/pkg/drift/gcs/gcs.go
@@ -64,24 +64,28 @@ func (c *Client) FilesWithName(ctx context.Context, bucket, filename string) ([]
 	return uris, nil
 }
 
-// DownloadFileIntoMemory fetches the file from GCS and reads it into memory.
+// DownloadAndUnmarshal fetches the file from GCS and decodes it using the
+// provided unmarshaller.
+//
 // TODO(dcreey): Handle race conditions https://github.com/abcxyz/guardian/issues/96.
-func (c *Client) DownloadFileIntoMemory(ctx context.Context, uri string) ([]byte, error) {
+func (c *Client) DownloadAndUnmarshal(ctx context.Context, uri string, unmarshaller func(r io.Reader) error) error {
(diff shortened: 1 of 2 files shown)


pkg/drift/gcs/gcs.go
```diff
@@ -64,24 +64,28 @@ func (c *Client) FilesWithName(ctx context.Context, bucket, filename string) ([]
 	return uris, nil
 }
 
-// DownloadFileIntoMemory fetches the file from GCS and reads it into memory.
+// DownloadAndUnmarshal fetches the file from GCS and decodes it using the
+// provided unmarshaller.
+//
 // TODO(dcreey): Handle race conditions https://github.com/abcxyz/guardian/issues/96.
-func (c *Client) DownloadFileIntoMemory(ctx context.Context, uri string) ([]byte, error) {
+func (c *Client) DownloadAndUnmarshal(ctx context.Context, uri string, unmarshaller func(r io.Reader) error) error {
```
dcreey, line 71: I think I prefer allowing any type to be passed here over passing a function to unmarshal. Then the api is pretty similar to the unmarshal api.

dcreey (approved):

pkg/drift/gcs/gcs.go
sethvargo: I didn't want to assume that the file contents would be JSON since this is a generic gcs client. This design supports XML, raw bytes, etc.

dcreey: NIT: Gotcha. I think its probably overkill since we are doing terraform drift detection tho? Might be yagni

drevell (approved):

PR #104, opened 2023-06-23
feat(drift): Allow filtering by project or folder

### LINKED ISSUE #100: [drift] Allow ignoring by folder or project
### TL;DR

We should enable users of the iam drift detection tool to ignore resources by Project ID, or Folder ID.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

dcreey:
Fixes #100 
Tested here: https://github.com/theteamatx/x-refinery-org/pull/158 (Created issue: https://github.com/theteamatx/x-refinery-org/issues/176)


.github/actions/iam-drift-detection/action.yml
```diff
@@ -50,7 +50,7 @@ runs:
         ORGANIZATION_ID: '${{ inputs.organization_id }}'
         GCS_BUCKET_LABEL: '${{ inputs.gcs_bucket_label }}'
       run: |
-        diff_snippet="$(.guardian/iam-drift-detection/detect_drift.sh "$ORGANIZATION_ID" "$GCS_BUCKET_LABEL")"
+        diff_snippet="$(go run ./cmd/guardian detect-iam-drift --organization-id="$ORGANIZATION_ID" --gcs-bucket-query="$GCS_BUCKET_LABEL")"
```
sethvargo: Go flags are short (single dash).

```suggestion
        diff_snippet="$(go run ./cmd/guardian detect-iam-drift -organization-id="$ORGANIZATION_ID" -
        gcs-bucket-query="$GCS_BUCKET_LABEL")"
```

Double dashes work, but are not required.

pkg/drift/driftignore.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+		} else { // Handle default case so we do not accidentally drop values.
+			filtered[k] = a
+		}
+	}
+
+	return filtered
+}
+
+func expandGraph(ignored *ignoredAssets, hierarchyGraph map[string]*hierarchyNode) (*ignoredAssets, error) {
```
sethvargo, line 68: I think we really need some unit coverage for this function and `foldersBeneath`.

dcreey: Oh yeah we do

.github/actions/iam-drift-detection/action.yml
dcreey: Done

pkg/drift/driftignore.go
dcreey: Moved to the assets package and added some tests.

pkg/drift/assets/assets.go
```diff
@@ -61,6 +61,17 @@ type HierarchyNode struct {
...
 }
 
+// HierarchyNodeWithChildren represents a node in the GCP Resource Hierarchy and all of its children.
+type HierarchyNodeWithChildren struct {
+	*HierarchyNode
+	ProjectIDs []string
+	FolderIDs  []string
+}
+
+type HierarchyGraph struct {
```
sethvargo, line 71: Doc this - what are the keys?

sethvargo (approved):

.github/actions/iam-drift-detection/action.yml
```diff
@@ -50,7 +50,9 @@ runs:
         ORGANIZATION_ID: '${{ inputs.organization_id }}'
         GCS_BUCKET_LABEL: '${{ inputs.gcs_bucket_label }}'
       run: |
-        diff_snippet="$(.guardian/iam-drift-detection/detect_drift.sh "$ORGANIZATION_ID" "$GCS_BUCKET_LABEL")"
+        driftignore_file="$(pwd)/.driftignore"
```
drevell, line 53: Maybe make this DRIFTIGNORE_FILE just for the sake of the convention of capitalizing env vars?

pkg/drift/driftignore.go
```diff
@@ -0,0 +1,143 @@
...
+	return &ignoredAssets{
+		iamAssets: ignored.iamAssets,
+		projects:  ignoredProjects,
+		folders:   ignoredFolders,
+	}, nil
+}
+
+// driftignore parses the driftignore file into a set.
+// Go doesn't implement set so we use a map of boolean values all set to true.
+func driftignore(ctx context.Context, fname string) (*ignoredAssets, error) {
```
drevell, line 89: Likely I'm missing some context, but how about using a YAML or JSON file here to avoid the complexity of writing a parser?

pkg/drift/driftignore.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func addListToSet(set map[string]struct{}, list []string) map[string]struct{} {
+	for _, v := range list {
+		set[v] = struct{}{}
+	}
+	return set
+}
+
+func mergeSets(setA, setB map[string]struct{}) map[string]struct{} {
```
drevell, line 138: Can we add a doc comment here saying that setA is mutated? The caller might be surprised that setA is an output parameter.

drevell (changes requested):

pkg/drift/assets/assets.go
dcreey: Done

pkg/drift/driftignore.go
dcreey: I think this is the likely direction this will go - however we don't yet have enough use cases to understand what is needed quite yet. I'll add a TODO to explore doing this but I want to delay until we have 1-2 real customers using it and we get real feedback (I will be the first customer).

I'll leave as is for now until we collect more feedback though. Let me know if you think this is appropriate?

pkg/drift/driftignore.go
drevell: Or even better, remove the return value. That will make it more clear that the argument is mutated in place.

pkg/drift/driftignore.go
drevell: FIne by me, as long as you've considered it.

dcreey: https://github.com/abcxyz/guardian/issues/105

.github/actions/iam-drift-detection/action.yml
dcreey: its not an env var though? Its just a bash variable. It would look weird to do that for `driftignore_file` and not for `diff_snippet`. wdyt?

pkg/drift/driftignore.go
dcreey: Done

.github/actions/iam-drift-detection/action.yml
drevell: Sorry, I misspoke, what I mean is that we conventionally capitalize shell variables, of which environment variables are only a subset. So maybe diff_snippet and driftignore_file should both be capitalized.

I don't have strong feeling on it, if you have a strong preference for it to stay this way, that's fine with me.

dcreey: Consulted the googs conventions go/shell-style#naming-conventions

Seems as though constants, env variables, and variables that are exported should be upper cased. Everything else should be lower. They do add that constants determined at runtime should still be uppercased. I think in this situation the `driftignore_file` qualifies as a constant so I uppercased it.

drevell: TIL of that style guide rule, thanks.

drevell (approved):

PR #107, opened 2023-06-27
feat(drift): Allow using project name in driftignore and show in output

dcreey:
Requested by @mmaraya in order to start using.

Tested: https://github.com/theteamatx/x-refinery-org/pull/158 (created issue: https://github.com/theteamatx/x-refinery-org/issues/181)

### DIFF
--- a/pkg/drift/assets/assets.go
+++ b/pkg/drift/assets/assets.go
@@ -269,6 +269,7 @@ func addFolderToGraph(graph map[string]*HierarchyNodeWithChildren, folder *Hiera
 	return nil
 }
 
+// AssetsByName returns a map of assets keyed by asset name.
 func AssetsByName(assetsByID map[string]*HierarchyNode) map[string]*HierarchyNode {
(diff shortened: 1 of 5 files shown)


pkg/drift/assets/assets.go
```diff
@@ -257,3 +268,22 @@ func addFolderToGraph(graph map[string]*HierarchyNodeWithChildren, folder *Hiera
...
+
+func AssetsByName(assetsByID map[string]*HierarchyNode) map[string]*HierarchyNode {
+	assetsByName := make(map[string]*HierarchyNode)
+	for _, a := range assetsByID {
+		assetsByName[a.Name] = a
+	}
+	return assetsByName
+}
+
+func Merge(assetsA, assetsB map[string]*HierarchyNode) map[string]*HierarchyNode {
```
sethvargo, line 280: Maybe document that assets in B take precedence/overwrite assets in A in the final result.

dcreey: Done

PR #113, opened 2023-06-28
feat: add child processing support

verbanicm:
Adds support for shelling out to child processes.

In future changes this will be used by other Guardian processes to interact with other binaries, such as Terraform, git, etc.


pkg/child/child.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, nil, 1, fmt.Errorf("failed to run command: %w", ctx.Err())
+	default:
+	}
+
+	c.cmd.Args = append(c.cmd.Args, c.args...)
+
+	// add small wait delay to kill subprocesses if context is canceled
+	// https://github.com/golang/go/issues/23019
+	// https://github.com/golang/go/issues/50436
+	c.cmd.WaitDelay = 2 * time.Second
```
sethvargo, line 79: Why not set this in `New`? It's a race by doing it in `Run`.

pkg/child/child.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := c.cmd.Wait(); err != nil {
+		return stdout.Bytes(), stderr.Bytes(), c.cmd.ProcessState.ExitCode(), fmt.Errorf("failed to run command: %w", err)
+	}
+
+	return stdout.Bytes(), stderr.Bytes(), c.cmd.ProcessState.ExitCode(), nil
+}
+
+// Command prints the command and args for the child process.
+func (c *Child) Command() string {
```
sethvargo, line 99: Is this used?

pkg/child/sys_linux.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
sethvargo, line 1: I think you need a build tag here

pkg/child/child.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if c.workingDir != "" {
+		c.cmd.Dir = c.workingDir
+	}
+
+	return c, nil
+}
+
+// Run executes the child process with the provided arguments.
+func (c *Child) Run(ctx context.Context) ([]byte, []byte, int, error) {
```
sethvargo, line 67: It's unclear what state we intend to modify on "child" between `New` and `Run`? Why not combine them?

verbanicm: Fair. Will update.

pkg/child/child.go
verbanicm: Future thinking potentially, but ill remove it for now.

pkg/child/child.go
verbanicm: Will adjust per later comment.

pkg/child/sys_linux.go
verbanicm: Ack

pkg/child/child.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Run executes a child process with the provided arguments.
+func Run(ctx context.Context, command string, args []string, workingDir string) ([]byte, []byte, int, error) {
+	path, err := exec.LookPath(command)
+	if err != nil {
+		return nil, nil, 1, fmt.Errorf("failed to locate command exec path: %w", err)
+	}
+
+	stdout := bytes.NewBuffer(nil)
+	stderr := bytes.NewBuffer(nil)
```
sethvargo, line 34: Style nit: `bytes.Buffer` has a stable zero value:

```go
var stdout, stderr bytes.Buffer
```

pkg/child/child.go
```diff
@@ -0,0 +1,68 @@
...
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+	cmd.Args = append(cmd.Args, args...)
+
+	// add small wait delay to kill subprocesses if context is canceled
+	// https://github.com/golang/go/issues/23019
+	// https://github.com/golang/go/issues/50436
+	cmd.WaitDelay = 2 * time.Second
+
+	select {
```
sethvargo, line 53: Nit: comment. I'm assuming this is to check if the context has been cancelled before the command actually starts? If so, I'm not sure `cmd.ProcessState.ExitCode()` will be defined - you might need to check if `cmd.ProcessState != nil` first.

pkg/child/child_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
sethvargo, line 113: Can you add a test where the context is cancelled _before_ the command even starts?

pkg/child/child.go
verbanicm: Removed this as cmd.Start() handles this case...see test.

pkg/child/child_test.go
verbanicm: Done.

pkg/child/child.go
verbanicm: Done

sethvargo (approved):

PR #114, opened 2023-06-30
feat: add packages for writing core guardian logic in go

verbanicm:
Create a set of packages for building out the core Guardian functionality, plan, apply, etc.

Created interfaces and implemented them to allow for mocking in testing that will be added later for commands, e.g. plan, apply, etc.


pkg/child/child.go
```diff
@@ -20,11 +20,22 @@ import (
...
 	"fmt"
 	"os/exec"
+	"strings"
 	"time"
 )
 
+var _ Runner = (*ChildRunner)(nil)
+
+// Runner defines the functionality for running child processes.
+type Runner interface {
```
verbanicm, line 30: Added this interface to allow for mocking, testing.

PR #116, opened 2023-07-05
feat: add terraform go interface

verbanicm:
Create a Terraform interface to define the available commands. This also allows for easier mocking for testing in the future.

Small change to child to make workingDir first argument after context.


pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+	childArgs = append(childArgs, args...)
+
+	if file != "" {
+		childArgs = append(childArgs, fmt.Sprintf("-out=%s", file))
+	}
+
+	return child.Run(ctx, workingDir, "terraform", childArgs) //nolint:wrapcheck
+}
+
+// Apply runs the Terraform plan command.
```
yolocs, line 91:
```suggestion
// Apply runs the Terraform apply command.
```

pkg/terraform/terraform_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	cases := []struct {
+		name string
+		dir  string
+		exp  []string
+		err  string
+	}{
+		{
+			name: "success_has_backend",
+			dir:  "../../terraform",
```
yolocs, line 106:
```suggestion
			dir:  "../../terraform", // depend on test data in [REPO_ROOT]/terraform
```

pkg/terraform/terraform_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			output := FormatOutputForGitHubDiff(tc.content)
+			if diff := cmp.Diff(output, tc.exp); diff != "" {
```
yolocs, line 88: nit: usually for primitive types, we just use `==` to compare. I'm not sure if cmp.Diff gives you better diff output if values under comparison are strings.

pkg/terraform/terraform_mock.go
```diff
@@ -0,0 +1,83 @@
...
+func (m *MockTerraformClient) Validate(ctx context.Context, workingDir string, args ...string) ([]byte, []byte, int, error) {
+	if m.ValidateResponse != nil {
+		return m.ValidateResponse.Stdout, m.ValidateResponse.Stderr, m.ValidateResponse.ExitCode, m.ValidateResponse.Err
+	}
+	return []byte{}, []byte{}, 0, nil
+}
+
+func (m *MockTerraformClient) Plan(ctx context.Context, workingDir, file string, args ...string) ([]byte, []byte, int, error) {
+	if m.PlanResponse != nil {
+		fmt.Printf("Plan response: %+v", m.PlanResponse)
```
yolocs, line 58: Remove?

yolocs (commented):
nits

pkg/terraform/terraform_test.go
verbanicm: These are both `[]string` (`dirs` is sorted so output should be consistent).

I can use `strings.Join` instead and use `!=` ?

yolocs: On line 30, `exp` is of type `string`?

yolocs (approved):

pkg/terraform/terraform_test.go
verbanicm: Ah, was looking in wrong spot. Fixed.

pkg/terraform/schema.go
```diff
@@ -0,0 +1,91 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package terraform
+
+import "github.com/hashicorp/hcl/v2"
+
+// RootSchema is the Terraform root schema.
+var RootSchema = &hcl.BodySchema{
```
drevell, line 20: Just checking: are we sure that this isn't already in library somewhere that we can import?

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+// NewTerraformClient creates a new Terraform client.
+func NewTerraformClient() *TerraformClient {
+	return &TerraformClient{}
+}
+
+// Init runs the terraform init command.
+func (t *TerraformClient) Init(ctx context.Context, workingDir string, args ...string) ([]byte, []byte, int, error) {
+	childArgs := []string{"init"}
+	childArgs = append(childArgs, args...)
+	return child.Run(ctx, workingDir, "terraform", childArgs) //nolint:wrapcheck
```
drevell, line 57: If you want to, you could make this a one-liner:

```
return child.Run(ctx, workingDir, "terraform", append([]string{"init"}, childArgs...))
```

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+	if err := filepath.WalkDir(rootDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return fmt.Errorf("failed to walk directory %s: %w", path, err)
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		matched, err := filepath.Match("*.tf", filepath.Base(path))
```
drevell, line 116: You might like [filepath.Ext](https://pkg.go.dev/path/filepath#Ext) to simplify this a bit

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+		matches[filepath.Dir(path)] = struct{}{}
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to find files: %w", err)
+	}
+
+	dirs := []string{}
+
+	for dir := range matches {
```
drevell, line 143: https://pkg.go.dev/golang.org/x/exp/maps#Keys would be good for this

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+	sort.Strings(dirs)
+
+	return dirs, nil
+}
+
+// hasBackendConfig tests a Terraform config file for the existence of a backend block.
+func hasBackendConfig(path string) (bool, hcl.Diagnostics, error) {
+	var diags hcl.Diagnostics
+
+	if _, err := os.Stat(path); err != nil {
```
drevell, line 156: This Stat() call is arguably redundant. We should be checking the result of ParseHCLFile anyway, which would include the case for nonexistent file.

https://pkg.go.dev/github.com/hashicorp/hcl/v2/hclparse#Parser.ParseHCLFile says `An error diagnostic is returned if the given file cannot be read.` Check out https://github.com/hashicorp/hcl/blob/v2.17.0/hclparse/parser.go#L78 to see what the error looks like.

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	return false, diags, nil
+}
+
+// FormatOutputForGitHubDiff formats the Terraform diff output for use with
+// GitHub diff markdown formatting.
+func FormatOutputForGitHubDiff(content string) string {
+	changed := regexp.MustCompile("(?m)^([\t ]*)([~])")
```
drevell, line 183: I'd suggest moving the MustCompile calls into a global scope, like:

```
var (
  changed = regexp.MustCompile(...)
  createReplace = regexp.MustCompile(...)
)

func FormatOutputForGitHubDiff(...
```

The advantages are:
 - The regexps are tested "for free"; since global variables are initialized on program start, they're implicitly tested by any test that touches this package
 - Just a little cleaner to compile once and reuse rather than compile repeatedly
 - Stylistically consistent with abcxyz

pkg/terraform/terraform_mock.go
```diff
@@ -0,0 +1,81 @@
...
+	ShowResponse     *MockTerraformResponse
+
+	EntrypointDirs []string
+}
+
+func (m *MockTerraformClient) Init(ctx context.Context, workingDir string, args ...string) ([]byte, []byte, int, error) {
+	if m.InitResponse != nil {
+		return m.InitResponse.Stdout, m.InitResponse.Stderr, m.InitResponse.ExitCode, m.InitResponse.Err
+	}
+	return []byte{}, []byte{}, 0, nil
```
drevell, line 45: We usually prefer to return nil rather than empty slice. https://google.github.io/styleguide/go/decisions#nil-slices

pkg/terraform/terraform.go
drevell, line 183: It might be helpful to document this regex with comments like:

```
regexp.MustCompile(
  "(?m)" + // enable multi-line mode
  " .... "  + // tildes are special because foo
```

pkg/terraform/terraform_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			tfclient := NewTerraformClient()
```
drevell, line 124: tfclient -> tfClient

pkg/terraform/terraform_test.go
```diff
@@ -0,0 +1,135 @@
...
+			dirs, err := tfclient.GetEntrypointDirectories(tc.dir)
+			if diff := testutil.DiffErrString(err, tc.err); diff != "" {
+				t.Errorf("unexpected error: %v", diff)
```
drevell, line 127: You can just drop the return value from DiffErrString into t.Error without wrapping it. `t.Error(diff)`

drevell (changes requested):

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+// TerraformClient implements the Terraform interface.
+type TerraformClient struct{}
+
+// NewTerraformClient creates a new Terraform client.
+func NewTerraformClient() *TerraformClient {
+	return &TerraformClient{}
+}
+
+// Init runs the terraform init command.
+func (t *TerraformClient) Init(ctx context.Context, workingDir string, args ...string) ([]byte, []byte, int, error) {
```
sethvargo, line 54: I would expect "workingDir" to be a function of the terraform client, not a param to the call.

sethvargo, line 54: These return types are strange. I assume it's stdout, stderr, exit code, and any error?

It might be better to return a `(struct, error)`, and the structure holds stdout, stderr, and the return code. (throughout)

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,190 @@
...
+		content, _, d := terraformBlock.Body.PartialContent(TerraformSchema)
+		diags = append(diags, d...)
+
+		innerBlocks := content.Blocks.OfType("backend")
+		if len(innerBlocks) > 0 {
+			return true, diags, nil
+		}
+	}
+
+	return false, diags, nil
```
sethvargo, line 177: Someone should build xpath, but for HCL...

pkg/terraform/terraform.go
verbanicm: Should I adopt this in `child.go` instead of here?

pkg/terraform/terraform.go
verbanicm: I am not sure I like this as i find it hard to read.

pkg/terraform/terraform.go
sethvargo: Yea probably

pkg/child/child.go
```diff
@@ -36,26 +50,38 @@ func Run(ctx context.Context, command string, args []string, workingDir string)
 	setSysProcAttr(cmd)
 	setCancel(cmd)
 
-	if workingDir != "" {
-		cmd.Dir = workingDir
+	if cfg.WorkingDir != "" {
+		cmd.Dir = cfg.WorkingDir
```
sethvargo, line 54: Style nit: 

```suggestion
	if v := cfg.WorkingDir; v != "" {
		cmd.Dir = v
```

pkg/child/child.go
```diff
@@ -23,11 +23,25 @@ import (
...
+type RunConfig struct {
+	WorkingDir string
+	Command    string
+	Args       []string
+}
+
+// RunResult is the response from a run operation.
+type RunResult struct {
+	Stdout   []byte
+	Stderr   []byte
```
sethvargo, line 36: Two questions:

1. Does it make sense to return bytes or string? 
2. Why not return an `io.Reader`?

The answer kinda depends on how the output is being used, but I'm a little bit worried that we're going to be loading mb (maybe even gb if debug logging is on) into memory.

drevell (approved):

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+var (
+	tildeChanged = regexp.MustCompile(
+		"(?m)" + // enable multi-line mode
+			"^([\t ]*)" + // only match tilde at start of line, can lead with tabs or spaces
+			"([~])") // tilde represents changes and needs switched to exclamation for git diff
```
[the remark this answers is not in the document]
verbanicm: appreciate the feedback, much more clear

pkg/terraform/terraform.go
```diff
@@ -0,0 +1,265 @@
...
+		innerBlocks := content.Blocks.OfType("backend")
+		if len(innerBlocks) > 0 {
+			return true, diags, nil
+		}
+	}
+
+	return false, diags, nil
+}
+
+var (
```
sethvargo, line 246: Nit: these are usually at the top

PR #117, opened 2023-07-06
refactor(drift): Parallelize drift ignore

### LINKED ISSUE #101: [drift] Implement multithreading for IAM fetching
### TL;DR

Implementing multithreading for the IAM fetching of folders and projects will significantly improve the performance of the tool.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

dcreey:
Fixes #101

### DIFF
--- a/pkg/commands/drift/drift.go
+++ b/pkg/commands/drift/drift.go
@@ -187,10 +187,11 @@ func actualGCPIAM(
 	})
 	for _, f := range foldersByID {
 		folderID := f.ID
+		folderName := f.Name
 		g.Go(func() error {
 			fIAM, err := client.FolderIAM(ctx, folderID)
 			if err != nil {
-				return fmt.Errorf("failed to get folder IAM for folder with ID '%s' and name '%s': %w", f.ID, f.Name, err)
+				return fmt.Errorf("failed to get folder IAM for folder with ID '%s' and name '%s': %w", folderID, folderName, err)
 			}
 			iamChan <- fIAM
 			return nil
@@ -198,10 +199,11 @@ func actualGCPIAM(
 	}
 	for _, p := range projectsByID {
 		projectID := p.ID
+		projectName := p.Name
 		g.Go(func() error {
 			pIAM, err := client.ProjectIAM(ctx, projectID)
 			if err != nil {
-				return fmt.Errorf("failed to get project IAM for project with ID '%s' and name '%s': %w", p.ID, p.Name, err)
+				return fmt.Errorf("failed to get project IAM for project with ID '%s' and name '%s': %w", projectID, projectName, err)
 			}
 			iamChan <- pIAM
 			return nil
--- a/pkg/commands/drift/drift.go
+++ b/pkg/commands/drift/drift.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/abcxyz/pkg/logging"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/abcxyz/guardian/pkg/commands/drift/assets"
 	"github.com/abcxyz/guardian/pkg/commands/drift/iam"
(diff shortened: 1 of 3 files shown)


pkg/commands/drift/drift.go
```diff
@@ -79,21 +93,51 @@ func Process(ctx context.Context, organizationID, bucketQuery, driftignoreFile s
...
 
-	gcpIAM, err := actualGCPIAM(ctx, organizationID, foldersByID, projectsByID)
+	g, gCtx := errgroup.WithContext(ctx)
+
+	logger.Debugw("fetching iam for org, folders and projects",
+		"number_of_folders", len(folders),
+		"number_of_projects", len(projects))
+	gcpIAMCount := 1 + len(foldersByID) + len(projectsByID)
+	gcpIAMChan := make(chan []*iam.AssetIAM, gcpIAMCount)
+	err = actualGCPIAM(gCtx, g, organizationID, foldersByID, projectsByID, gcpIAMChan)
```
drevell, line 103: I'm curious why this function uses a different concurrency style than the previous (this one uses channels, the other one assigns to a local variable from a goroutine). The other one seemed cleaner to me, since it avoids all the complexity of channel buffering and such, but that's just my opinion. Could `gcpIAM` and `tfIAM` just be return values rather than being read from a channel? You always have the option of creating new/separate nested errgroups inside terraformStateIAM and actualGCPIAM.

Likely I'm just missing something, but it's worth discussing.

pkg/commands/drift/drift.go
```diff
@@ -79,21 +93,51 @@ func Process(ctx context.Context, organizationID, bucketQuery, driftignoreFile s
 		return nil, fmt.Errorf("failed to expand graph for ignored assets: %w", err)
 	}
 
-	gcpIAM, err := actualGCPIAM(ctx, organizationID, foldersByID, projectsByID)
+	g, gCtx := errgroup.WithContext(ctx)
```
drevell, line 96: I don't know how this will be used in practice, but should we be worried about issuing too many concurrent API requests and getting throttled? Perhaps use https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit.

drevell (changes requested):

pkg/commands/drift/drift.go
dcreey: There were a few things driving my design here:

1) I wanted to fully parallelize the GCP IAM and Terraform IAM operations so they happened concurrently. This meant that I couldn't fully encapsulate the waitGroup/errorGroup in their methods. (I wish there was a `Promise.All()` equivalent for combining waits from multiple goroutines/waitGroups).
2) Use of channels here is required because we have concurrent writes to the same collection (map). I could instead use a mutex and lock before mutating the collection. IMO this might be a bit cleaner since it avoids the translation back from channels to the data format we want.
3) We didn't have to use a mutex on the first 3 operations since they aren't mutating the same collections. Its 1 operation per expected output.

I could take the hit so that we don't parallelize the TF IAM and GCP IAM ops - I do think this might be a little cleaner.

pkg/commands/drift/drift.go
dcreey: Yeah. Good idea. Was waiting to get feedback on errGroup vs WaitGroup vs something else like https://github.com/sourcegraph/conc

drevell: I'd love to start using conc, maybe this is a good time. Though I imagine @sethvargo will have strong feelings here and he's OOO. If you want to take a stab at using conc, I'd approve it.

The safe choice in my opinion would be to stick with errgroup for now. I'd choose it over WaitGroup because it saves you from managing errors yourself.

pkg/commands/drift/drift.go
drevell: Those reasons sound good to me. I'm fine with leaving it as is, or rewriting per your idea in your last paragraph.

pkg/commands/drift/drift.go
dcreey: Using conc: https://github.com/abcxyz/guardian/pull/118/files

drevell: Dropped a comment on that (#118) suggesting a simplification. The conc approach seems good to me.

sethvargo: We have a thing that does this and is safer (both type and error safe): https://github.com/abcxyz/pkg/tree/main/worker

PR #118, opened 2023-07-07
refactor(drift): Parallelize requests using conc

dcreey:


pkg/commands/drift/drift.go
```diff
@@ -129,34 +159,58 @@ func actualGCPIAM(
...
-	oIAM, err := client.OrganizationIAM(ctx, organizationID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get organization IAM for ID '%s': %w", organizationID, err)
-	}
-	for _, i := range oIAM {
-		m[URI(i, organizationID, foldersByID, projectsByID)] = i
-	}
-	for _, f := range foldersByID {
-		fIAM, err := client.FolderIAM(ctx, f.ID)
+	p := pool.New().
```
drevell, line 162: If we used NewWithResults / ResultPool, could we avoid the need for channels?

drevell (changes requested):

drevell:
Thanks for trying this out, this is fun.

pkg/commands/drift/drift.go

drevell (approved):

drevell:
> Should note that using the pool comes with some overhead.
> 
> ```
> // Pool is efficient, but not zero cost. It should not be used for very short
> // tasks. Startup and teardown come with an overhead of around 1µs, and each
> // task has an overhead of around 300ns.
> ```
> 
> https://github.com/sourcegraph/conc/blob/main/pool/pool.go#L27C1-L29C41
> 
> This shouldn't be an issue for something like iam drift detection - but this could be more problematic if you want to write something high performance.

Good to know, thanks for the education.

PR #119, opened 2023-07-10
refactor(drift): Switch to using abcxyz worker for parallelization

dcreey:


pkg/commands/drift/drift.go
```diff
@@ -44,34 +44,40 @@ func Process(ctx context.Context, organizationID, bucketQuery, driftignoreFile s
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize assets client: %w", err)
 	}
+	w := worker.New[*worker.Void](3)
```
sethvargo, line 47: We should make the concurrency configurable (can be a follow-up PR).

sethvargo (dismissed):

pkg/commands/drift/drift.go

sethvargo (approved):

sethvargo (approved):

PR #120, opened 2023-07-11
fix: address previous pr comments

verbanicm:
Address comments from #116

### DIFF
--- a/pkg/child/child_test.go
+++ b/pkg/child/child_test.go
@@ -15,13 +15,14 @@
 package child
 
 import (
+	"bytes"
 	"context"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/testutil"
-	"github.com/google/go-cmp/cmp"
 )
 
 func TestChild_Run(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/child/child_test.go
```diff
@@ -132,8 +152,25 @@ func TestChild_Run_Cancel(t *testing.T) {
 			if diff := testutil.DiffErrString(err, tc.err); diff != "" {
 				t.Errorf(diff)
 			}
-			if diff := cmp.Diff(result, tc.exp); diff != "" {
-				t.Errorf("result differed from expected, (-got,+want): %s", diff)
+
+			stdout := make([]byte, 0)
+			stderr := make([]byte, 0)
+
+			if _, err := result.Stdout.Read(stdout); err != nil {
```
drevell, line 159: Maybe https://pkg.go.dev/io#ReadAll is the way to go here. I don't think Read() guarantees that all bytes will be read (hence the `n int` return value).

drevell (approved):

pkg/child/child_test.go
verbanicm: That makes way more sense. Thank you!

drevell (approved):

sethvargo (approved):

PR #121, opened 2023-07-11
feat: add git package

verbanicm:


pkg/git/git.go
```diff
@@ -0,0 +1,76 @@
...
+// NewGitClient creates a new Terraform client.
+func NewGitClient() *GitClient {
+	return &GitClient{}
+}
+
+// DiffDirs runs a git diff between two revisions and returns the sorted ist of directories with changes.
+func (g *GitClient) DiffDirs(ctx context.Context, workingDir, baseRef, headRef string) ([]string, error) {
+	matches := make(map[string]struct{})
+
+	result, err := child.Run(ctx, &child.RunConfig{
```
verbanicm, line 51: I think adding an interface for child might make sense to we can mock commands to provide better test coverage in the future.

Thoughts?

pkg/git/git.go
```diff
@@ -0,0 +1,76 @@
...
+		if len(line) > 0 {
+			matches[filepath.Dir(line)] = struct{}{}
+		}
+	}
+
+	dirs := maps.Keys(matches)
+
+	sort.Strings(dirs)
+
+	return dirs, nil
```
yolocs, line 75: Would recommend making a helper func of it and add basic unit test.

pkg/git/git.go
```diff
@@ -0,0 +1,76 @@
...
+// GitClient implements the git interface.
+type GitClient struct{}
+
+// NewGitClient creates a new Terraform client.
+func NewGitClient() *GitClient {
+	return &GitClient{}
+}
+
+// DiffDirs runs a git diff between two revisions and returns the sorted ist of directories with changes.
+func (g *GitClient) DiffDirs(ctx context.Context, workingDir, baseRef, headRef string) ([]string, error) {
```
sethvargo, line 48: I'm having nightmares about the order of arguments for `ln`... 3 string properties in a row feels like a recipe for misuse. Should we accept a struct instead?

verbanicm: I can move the working dir to be a property of the struct? Then its just 2 strings?

pkg/git/git.go
verbanicm: SGTM, this was my fallback instead of an interface. Ill update.

pkg/git/git.go
verbanicm: moved

pkg/git/git_test.go
```diff
@@ -0,0 +1,70 @@
...
+	t.Parallel()
+
+	cases := []struct {
+		name  string
+		value string
+		exp   []string
+	}{
+		{
+			name:  "success",
+			value: "first/test.txt\nsecond/test.txt\nthird/test.txt",
```
yolocs, line 33: nit: you could do
```suggestion
			value: `first/test.txt
second/test.txt
third/test.txt`,
```

yolocs (dismissed):

sethvargo (approved):

PR #123, opened 2023-07-12
feat: add storage package

verbanicm:
@dcreey needs this for iam-drift to create single shared package.

### DIFF
--- a/pkg/storage/google_cloud_storage.go
+++ b/pkg/storage/google_cloud_storage.go
@@ -15,7 +15,6 @@
 package storage
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -25,6 +24,8 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/googleapis/gax-go/v2"
 )
+
+const MiB = 1_048_576 // 1 MiB
 
 var _ Storage = (*GoogleCloudStorage)(nil)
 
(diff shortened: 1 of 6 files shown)


pkg/storage/google_cloud_storage.go
```diff
@@ -0,0 +1,213 @@
...
+	defer cancel()
+
+	attrs, err := o.Attrs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get object metadata: %w", err)
+	}
+	return attrs.Metadata, nil
+}
+
+func (s *GoogleCloudStorage) DeleteObject(ctx context.Context, bucket, name string, ignoreNotFound bool) error {
```
sethvargo, line 167: ignoreNotFound feels like a weird API; why would we _not_ want to ignore it? The caller says "ensure object FOO is deleted". If it doesn't exist, the invariant is satisfied.

pkg/storage/google_cloud_storage.go
```diff
@@ -0,0 +1,213 @@
...
+	defer cancel()
+
+	writer := o.NewWriter(ctx)
+
+	if contentType != "" {
+		writer.ContentType = contentType
+	}
+
+	if metadata != nil {
+		writer.Metadata = metadata
```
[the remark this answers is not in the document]
verbanicm: I feel like that was javascript specific, the docs mention this property is for user provided metadata, not all metadata. Made the change anyway.

pkg/storage/google_cloud_storage.go
```diff
@@ -0,0 +1,218 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/googleapis/gax-go/v2"
+)
+
+const MiB = 1_048_576 // 1 MiB
```
sethvargo, line 28: Nit: `1 << 20` might be more explicit

pkg/storage/google_cloud_storage.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (s *GoogleCloudStorage) UploadObject(ctx context.Context, bucket, name string, contents []byte, opts ...UploadOption) error {
+	cfg := makeUploadConfig(len(contents), opts)
+
+	o, ctx, cancel := s.objectHandleWithRetries(ctx, bucket, name)
+	defer cancel()
+
+	writer := o.If(storage.Conditions{DoesNotExist: true}).NewWriter(ctx)
+	defer writer.Close()
```
sethvargo, line 123: You still want to check the error return value here though.

pkg/storage/google_cloud_storage.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	if _, err := writer.Write(contents); err != nil {
+		return fmt.Errorf("failed to write data: %w", err)
+	}
+
+	return nil
+}
+
+func (s *GoogleCloudStorage) DownloadObject(ctx context.Context, bucket, name string) (io.ReadCloser, error) {
```
sethvargo, line 149: Doc: it's up to the caller to close the reader

verbanicm: Should this go in the interface doc or the implementation....or both?

PR #129, opened 2023-07-13
feat: add github package

verbanicm:


pkg/github/github.go
```diff
@@ -0,0 +1,195 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/google/go-github/v53/github"
+	"github.com/sethvargo/go-retry"
+	"golang.org/x/oauth2"
+)
+
+var ignoredStatusCodes = []int{400, 401, 403, 404, 422}
```
sethvargo, line 28: I'd suggest making this a `map[int]struct{}`

pkg/github/github.go
```diff
@@ -0,0 +1,195 @@
...
+	for _, opt := range opts {
+		if opt != nil {
+			cfg = opt(cfg)
+		}
+	}
+
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tc := oauth2.NewClient(ctx, ts)
```
sethvargo, line 69: Is there an SDK function for "source the creds from the env"? That would let us support private key auth and token auth easily.

pkg/github/github.go
```diff
@@ -0,0 +1,195 @@
...
+	g := &GitHubClient{
+		cfg:    cfg,
+		client: client,
+	}
+
+	return g
+}
+
+// CreatePullRequestComment creates a pull request comment.
+func (g *GitHubClient) CreatePullRequestComment(ctx context.Context, owner, repo string, number int, body string) (int64, error) {
```
sethvargo, line 82: It feels weird that this only returns an ID?

pkg/github/github.go
```diff
@@ -0,0 +1,195 @@
...
+	return nil
+}
+
+// ListPullRequestComments lists existing comments for a pull request.
+func (g *GitHubClient) ListPullRequestComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*github.IssueComment, *github.Response, error) {
+	backoff := retry.NewFibonacci(g.cfg.initialRetryDelay)
+	backoff = retry.WithMaxRetries(g.cfg.maxRetries, backoff)
+	backoff = retry.WithCappedDuration(g.cfg.maxRetryDelay, backoff)
+
+	commentsResponse := make([]*github.IssueComment, 0)
```
sethvargo, line 164: Nit: 
```suggestion
	var commentsResponse []*github.IssueComment
```

bradegler (approved):

pkg/github/github.go
verbanicm: Do you mean so we could auth as a user or as an app? It seems there are two different methods for this: https://github.com/google/go-github#authentication

sethvargo: Yea, I'm thinking we might auth as an app in the future.

verbanicm: We have: https://github.com/abcxyz/pkg/blob/main/githubapp/githubapp.go

We can wrap this, but I think we'll need to accept an HTTP Client and build a wrapper based on inputs to generate the HTTP client we want to use...doesnt seem like it exists.

Ill add an issue for this.

verbanicm: Created #130 and added comment inline.

sethvargo (approved):

PR #133, opened 2023-07-14
fix: updates per design discussion offline

verbanicm:
Accept a writer instead of a reader for child to allow for greater flexibility.

- Refactor child and terraform
- Fix other bugs in git package


pkg/storage/options.go
```diff
@@ -52,6 +53,14 @@ func WithContentType(contentType string) UploadOption {
 	}
 }
 
+// WithAllowOverwrite sets the overwrite flag to allow overwriting the destination object.
+func WithAllowOverwrite(allowOverwrite bool) UploadOption {
```
verbanicm, line 57: This is needed because multiple plan runs will need to overwrite the existing plan file in the GCS bucket.

pkg/terraform/init_test.go
```diff
@@ -0,0 +1,95 @@
...
+package terraform
+
+import (
+	"testing"
+
+	"github.com/abcxyz/guardian/pkg/util"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestTerraform_initArgsFromOptions(t *testing.T) {
```
drevell, line 24: Would this be more conventional named TestInitArgsFromOptions?

pkg/util/util.go
```diff
@@ -0,0 +1,51 @@
...
+				break
+			}
+		}
+	}
+
+	result := []string{}
+
+	for r := range intersection {
+		result = append(result, r)
+	}
```
drevell, line 46: maps.Keys() can do this for you

drevell (changes requested):

sethvargo (approved):
This feels better

drevell (approved):

PR #134, opened 2023-07-14
feat(drift): Create issues from go

### LINKED ISSUE #111: [drift] Comment on open issue instead of closing and re-creating
### TL;DR

The current process for iam-drift-detection is to close any open issue and create a new one on every execution. Instead we should hide any previous comments and post a new comment. This will allow users to properly address issues, add comments, audit log, etc.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

dcreey:
Also fixes #111 

Tested: https://github.com/theteamatx/x-refinery-org/issues/206

### DIFF
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -17,7 +17,6 @@ package drift
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
@@ -107,14 +106,14 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 		Name:    "github-owner",
 		Target:  &c.flagGitHubOwner,
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
-		Default: os.Getenv("GITHUB_REPOSITORY_OWNER"),
+		Default: "",
 	})
 	f.StringVar(&cli.StringVar{
 		Name:    "github-repo",
 		Target:  &c.flagGitHubRepo,
 		Example: "guardian",
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
-		Default: os.Getenv("GITHUB_REPOSITORY"),
+		Default: "",
 	})
 	f.StringSliceVar(&cli.StringSliceVar{
 		Name:    "github-issue-assignees",
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -100,6 +100,7 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 		Name:    "github-token",
 		Target:  &c.flagGitHubToken,
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		EnvVar:  "GITHUB_TOKEN",
 		Default: "",
 	})
 	f.StringVar(&cli.StringVar{
@@ -158,12 +159,12 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
 
 	iamDriftDetector, err := NewIAMDriftDetector(ctx, c.flagOrganizationID, c.flagMaxConcurrentRequests)
 	if err != nil {
-		return fmt.Errorf("failed to create iam drift detector %w", err)
+		return fmt.Errorf("failed to create iam drift detector: %w", err)
 	}
 
 	iamDiff, err := iamDriftDetector.DetectDrift(ctx, c.flagGCSBucketQuery, c.flagDriftignoreFile)
 	if err != nil {
-		return fmt.Errorf("failed to detect drift %w", err)
+		return fmt.Errorf("failed to detect drift: %w", err)
 	}
 
 	changesDetected := len(iamDiff.ClickOpsChanges) > 0 || len(iamDiff.MissingTerraformChanges) > 0
@@ -177,11 +178,11 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
 	}
 	if changesDetected {
 		if err = createOrUpdateIssue(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, m); err != nil {
-			return fmt.Errorf("failed to manage GitHub Issue %w", err)
+			return fmt.Errorf("failed to create or update GitHub Issue: %w", err)
 		}
 	} else {
 		if err = closeIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueLabels); err != nil {
-			return fmt.Errorf("failed to close GitHub Issues %w", err)
+			return fmt.Errorf("failed to close GitHub Issues: %w", err)
 		}
 	}
 
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -17,8 +17,6 @@ package drift
 import (
 	"context"
 	"fmt"
-	"sort"
-	"strings"
 
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
@@ -38,6 +36,12 @@ type DetectIamDriftCommand struct {
 	flagGCSBucketQuery        string
 	flagDriftignoreFile       string
 	flagMaxConcurrentRequests int64
+	flagSkipGitHubIssue       bool
+	flagGitHubToken           string
+	flagGitHubOwner           string
+	flagGitHubRepo            string
+	flagGitHubLabels          []string
+	flagGitHubAssignees       []string
 }
 
 func (c *DetectIamDriftCommand) Desc() string {
@@ -85,6 +89,48 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 		Usage:   `The maximum number of concurrent requests allowed at any time to GCP.`,
 		Default: 10,
 	})
+	f.BoolVar(&cli.BoolVar{
+		Name:    "skip-github-issue",
+		Target:  &c.flagSkipGitHubIssue,
+		Example: "true",
+		Usage:   `Whether or not to create a GitHub Issue when a drift is detected.`,
+		Default: false,
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "gh-token",
+		Target:  &c.flagGitHubToken,
+		Example: "...",
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "gh-owner",
+		Target:  &c.flagGitHubOwner,
+		Example: "...",
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "gh-repo",
+		Target:  &c.flagGitHubRepo,
+		Example: "...",
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
+	})
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "gh-assignees",
+		Target:  &c.flagGitHubAssignees,
+		Example: "dcreey",
+		Usage:   `The assignees to assign to for any created GitHub Issues.`,
+		Default: []string{},
+	})
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "gh-labels",
+		Target:  &c.flagGitHubLabels,
+		Example: "guardian-iam-drift",
+		Usage:   `The labels to use on any created GitHub Issues.`,
+		Default: []string{"guardian-iam-drift"},
+	})
 
 	return set
 }
@@ -122,20 +168,15 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to detect drift %w", err)
 	}
 
-	// Output to stdout to mimic bash script for now.
-	// TODO(dcreey): Determine cleaner API that aligns with using the cli tool.
-	if len(iamDiff.ClickOpsChanges) > 0 {
-		uris := keys(iamDiff.ClickOpsChanges)
-		sort.Strings(uris)
-		c.Outf("Found Click Ops Changes \n> %s", strings.Join(uris, "\n> "))
-		if len(iamDiff.MissingTerraformChanges) > 0 {
-			c.Outf("\n\n")
-		}
+	if len(iamDiff.ClickOpsChanges) > 0 || len(iamDiff.MissingTerraformChanges) > 0 {
+		c.Outf(driftMessage(iamDiff))
 	}
-	if len(iamDiff.MissingTerraformChanges) > 0 {
-		uris := keys(iamDiff.MissingTerraformChanges)
-		sort.Strings(uris)
-		c.Outf("Found Missing Terraform Changes \n> %s", strings.Join(uris, "\n> "))
+
+	if !c.flagSkipGitHubIssue {
+		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubAssignees, c.flagGitHubLabels, iamDiff)
+		if err != nil {
+			return fmt.Errorf("failed to manage GitHub Issue %w", err)
+		}
 	}
 
 	return nil
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -166,14 +166,22 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to detect drift %w", err)
 	}
 
-	if len(iamDiff.ClickOpsChanges) > 0 || len(iamDiff.MissingTerraformChanges) > 0 {
-		c.Outf(driftMessage(iamDiff))
+	changesDetected := len(iamDiff.ClickOpsChanges) > 0 || len(iamDiff.MissingTerraformChanges) > 0
+	m := driftMessage(iamDiff)
+	if changesDetected {
+		c.Outf(m)
 	}
 
-	if !c.flagSkipGitHubIssue {
-		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, iamDiff)
-		if err != nil {
+	if c.flagSkipGitHubIssue {
+		return nil
+	}
+	if changesDetected {
+		if err = createOrUpdateIssue(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, m); err != nil {
 			return fmt.Errorf("failed to manage GitHub Issue %w", err)
+		}
+	} else {
+		if err = closeIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueLabels); err != nil {
+			return fmt.Errorf("failed to close GitHub Issues %w", err)
 		}
 	}
 
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -40,8 +40,8 @@ type DetectIamDriftCommand struct {
 	flagGitHubToken           string
 	flagGitHubOwner           string
 	flagGitHubRepo            string
-	flagGitHubLabels          []string
-	flagGitHubAssignees       []string
+	flagGitHubIssueLabels     []string
+	flagGitHubIssueAssignees  []string
 }
 
 func (c *DetectIamDriftCommand) Desc() string {
@@ -97,36 +97,34 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 		Default: false,
 	})
 	f.StringVar(&cli.StringVar{
-		Name:    "gh-token",
+		Name:    "github-token",
 		Target:  &c.flagGitHubToken,
-		Example: "...",
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
 		Default: "",
 	})
 	f.StringVar(&cli.StringVar{
-		Name:    "gh-owner",
+		Name:    "github-owner",
 		Target:  &c.flagGitHubOwner,
-		Example: "...",
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
 		Default: "",
 	})
 	f.StringVar(&cli.StringVar{
-		Name:    "gh-repo",
+		Name:    "github-repo",
 		Target:  &c.flagGitHubRepo,
-		Example: "...",
+		Example: "guardian",
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
 		Default: "",
 	})
 	f.StringSliceVar(&cli.StringSliceVar{
-		Name:    "gh-assignees",
-		Target:  &c.flagGitHubAssignees,
+		Name:    "github-issue-assignees",
+		Target:  &c.flagGitHubIssueAssignees,
 		Example: "dcreey",
 		Usage:   `The assignees to assign to for any created GitHub Issues.`,
 		Default: []string{},
 	})
 	f.StringSliceVar(&cli.StringSliceVar{
-		Name:    "gh-labels",
-		Target:  &c.flagGitHubLabels,
+		Name:    "github-issue-labels",
+		Target:  &c.flagGitHubIssueLabels,
 		Example: "guardian-iam-drift",
 		Usage:   `The labels to use on any created GitHub Issues.`,
 		Default: []string{"guardian-iam-drift"},
@@ -173,7 +171,7 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
 	}
 
 	if !c.flagSkipGitHubIssue {
-		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubAssignees, c.flagGitHubLabels, iamDiff)
+		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, iamDiff)
 		if err != nil {
 			return fmt.Errorf("failed to manage GitHub Issue %w", err)
 		}
--- a/pkg/commands/drift/command.go
+++ b/pkg/commands/drift/command.go
@@ -17,6 +17,7 @@ package drift
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
@@ -106,14 +107,14 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 		Name:    "github-owner",
 		Target:  &c.flagGitHubOwner,
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
-		Default: "",
+		Default: os.Getenv("GITHUB_REPOSITORY_OWNER"),
 	})
 	f.StringVar(&cli.StringVar{
 		Name:    "github-repo",
 		Target:  &c.flagGitHubRepo,
 		Example: "guardian",
 		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
-		Default: "",
+		Default: os.Getenv("GITHUB_REPOSITORY"),
 	})
 	f.StringSliceVar(&cli.StringSliceVar{
 		Name:    "github-issue-assignees",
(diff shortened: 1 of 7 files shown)


pkg/commands/drift/command.go
```diff
@@ -85,6 +89,48 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+		Name:    "skip-github-issue",
+		Target:  &c.flagSkipGitHubIssue,
+		Example: "true",
+		Usage:   `Whether or not to create a GitHub Issue when a drift is detected.`,
+		Default: false,
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "gh-token",
+		Target:  &c.flagGitHubToken,
+		Example: "...",
```
verbanicm, line 102: Remove this instead of ambiguous ... throughout.

pkg/commands/drift/command.go
```diff
@@ -85,6 +89,48 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
 	})
+	f.BoolVar(&cli.BoolVar{
+		Name:    "skip-github-issue",
+		Target:  &c.flagSkipGitHubIssue,
+		Example: "true",
+		Usage:   `Whether or not to create a GitHub Issue when a drift is detected.`,
+		Default: false,
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "gh-token",
```
verbanicm, line 100: Don't abbreviate gh, use github-*

pkg/github/github.go
```diff
@@ -171,13 +274,12 @@ func (g *GitHubClient) DeleteIssueComment(ctx context.Context, owner, repo strin
 }
 
 // ListIssueComments lists existing comments for an issue or pull request.
-func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*github.IssueComment, *github.Response, error) {
+func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*int64, error) {
```
verbanicm, line 277: I dont know if I agree with this, we may want to filter based on comment body or author, etc.

pkg/commands/drift/command.go
```diff
@@ -38,6 +36,12 @@ type DetectIamDriftCommand struct {
 	flagGCSBucketQuery        string
 	flagDriftignoreFile       string
 	flagMaxConcurrentRequests int64
+	flagSkipGitHubIssue       bool
+	flagGitHubToken           string
+	flagGitHubOwner           string
+	flagGitHubRepo            string
```
verbanicm, line 42: In my implementation of Plan that i haven't submitted yet, I am pulling these GH specific variables from the context using https://github.com/sethvargo/go-githubactions.

Ive been fighting myself on the best way to do this accept as flag or use the environment.

verbanicm: I think I prefer your way to using the context.

dcreey: Context/env var is weird to me. Especially since we are a cli tool there is no guarantee we will be run from a github repo/context with the right env var

pkg/commands/drift/command.go
dcreey: Done

pkg/commands/drift/command.go
dcreey: Done

pkg/github/github.go
dcreey: Perhaps we should add support for those when we need to? We can return a lightweight struct of the additional values when we determine what we need?

verbanicm: Can we just create the struct now and just have the ID? That way we dont have to change types, just add fields? (same for GetIssueComment and GetIssue also)

sethvargo: I agree with Mike

dcreey: Done

pkg/commands/drift/command.go
verbanicm: Hm, that might be true for the IAM drift, but plan/apply will be tightly coupled to running inside a pull request (creating comments, grouping outputs, etc).

pkg/github/github.go
```diff
@@ -93,13 +108,110 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
 	return g
 }
 
+type Issue struct {
```
verbanicm, line 111: Can you move these to the top, it feels strange placed in the middle of the file like this (both Issue and IssueComment)

pkg/github/github.go
```diff
@@ -171,13 +282,12 @@ func (g *GitHubClient) DeleteIssueComment(ctx context.Context, owner, repo strin
 }
 
 // ListIssueComments lists existing comments for an issue or pull request.
-func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*github.IssueComment, *github.Response, error) {
+func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*int64, error) {
```
verbanicm, line 285: This should return `[]*IssueComment`

dcreey: Done

pkg/github/github.go
dcreey: Done

pkg/commands/drift/command.go
dcreey: I think my preference would be to use a pattern like:

```
default := env.get('REPO')
f := flag.value('repo', default)
```

That way we can still use flags but we fallback to the environment. 

My NIT about https://github.com/sethvargo/go-githubactions is that this seems like a fairly heavyweight solution to just expose a couple of env variables. Additionally, I skimmed the README and it wasn't obvious to me that this was the right solution (improved documentation might help this).

pkg/github/github.go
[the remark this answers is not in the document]
verbanicm: Im thinking of a large pull request with lots of comments where we want to list all comments and delete only the previous ones for guardian. There COULD be >100. I have an immediate use case for it. I can add this when I implement if you want and we can leave it off for now.

verbanicm (approved):

pkg/commands/drift/command.go
```diff
@@ -85,6 +90,46 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+		Target:  &c.flagGitHubToken,
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "github-owner",
+		Target:  &c.flagGitHubOwner,
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: os.Getenv("GITHUB_REPOSITORY_OWNER"),
+	})
```
verbanicm, line 111: This is provided by the cli flags package:

```suggestion
	f.StringVar(&cli.StringVar{
		Name:    "github-owner",
		EnvVar:  "GITHUB_REPOSITORY_OWNER",
		Target:  &c.flagGitHubOwner,
		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
	})
```

pkg/commands/drift/command.go
```diff
@@ -85,6 +90,46 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+	f.StringVar(&cli.StringVar{
+		Name:    "github-repo",
+		Target:  &c.flagGitHubRepo,
+		Example: "guardian",
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: os.Getenv("GITHUB_REPOSITORY"),
```
verbanicm, line 117: This envvar is the full repo name: owner/repo not just the name.

dcreey: Discussed offline. Removed env var for time being.

pkg/commands/drift/command.go
```diff
@@ -122,20 +166,15 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
...
+		c.Outf(driftMessage(iamDiff))
 	}
-	if len(iamDiff.MissingTerraformChanges) > 0 {
-		uris := keys(iamDiff.MissingTerraformChanges)
-		sort.Strings(uris)
-		c.Outf("Found Missing Terraform Changes \n> %s", strings.Join(uris, "\n> "))
+
+	if !c.flagSkipGitHubIssue {
+		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, iamDiff)
+		if err != nil {
```
sethvargo, line 175:
```suggestion
		if err := createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, iamDiff); err != nil {
```

pkg/commands/drift/command.go
```diff
@@ -122,20 +166,15 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
...
+	if len(iamDiff.ClickOpsChanges) > 0 || len(iamDiff.MissingTerraformChanges) > 0 {
+		c.Outf(driftMessage(iamDiff))
 	}
-	if len(iamDiff.MissingTerraformChanges) > 0 {
-		uris := keys(iamDiff.MissingTerraformChanges)
-		sort.Strings(uris)
-		c.Outf("Found Missing Terraform Changes \n> %s", strings.Join(uris, "\n> "))
+
+	if !c.flagSkipGitHubIssue {
+		err = createUpdateOrCloseIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, iamDiff)
```
sethvargo, line 174: This seems to be a "junk drawer" function as evidenced by both it's name and the number of parameters. Does it make sense for this to be it's own struct with its own logic/encapuslation?

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+	"sort"
+	"strings"
+
+	"github.com/abcxyz/guardian/pkg/github"
+)
+
+const (
+	issueTitle = "IAM drift detected"
+	issueBody  = `We've detected a drift between your submitted IAM policies and actual
+		IAM policies.
```
sethvargo, line 29: Nit: these appear to be indented with tabs, which _will_ render in the GitHub PR.

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+		Please determine which parts are correct, and submit updated
+		terraform config and/or remove the extra policies.
+
+		Re-run drift detection manually once complete to verify all diffs are properly resolved.`
+)
+
+func createUpdateOrCloseIssues(ctx context.Context, token, owner, repo string, assignees, labels []string, drift *IAMDrift) error {
+	if len(labels) == 0 {
+		return fmt.Errorf("invalid argument - at least one 'label' must be provided")
+	}
```
sethvargo, line 42: Why?

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+
+		See the comment(s) below to see details of the drift
+
+		Please determine which parts are correct, and submit updated
+		terraform config and/or remove the extra policies.
+
+		Re-run drift detection manually once complete to verify all diffs are properly resolved.`
+)
+
+func createUpdateOrCloseIssues(ctx context.Context, token, owner, repo string, assignees, labels []string, drift *IAMDrift) error {
```
sethvargo, line 39: As I predicted in my previous comment, this function is doing wayyy to much. We need to break it down into smaller pieces.

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+	if len(labels) == 0 {
+		return fmt.Errorf("invalid argument - at least one 'label' must be provided")
+	}
+
+	changesDetected := len(drift.ClickOpsChanges) > 0 || len(drift.MissingTerraformChanges) > 0
```
sethvargo, line 44: Is this relevant? You already compute this in command.go

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+
+	issues, err := gh.ListIssues(ctx, owner, repo, labels, github.Open)
+	if err != nil {
+		return fmt.Errorf("failed to list GitHub issues for %s/%s: %w", owner, repo, err)
+	}
+	// Close all open issues matching the given labels if the drift is resolved.
+	if !changesDetected && len(issues) > 0 {
+		for _, issueToClose := range issues {
+			_, err = gh.CreateIssueComment(ctx, owner, repo, *issueToClose.Number, "Drift Resolved.")
+			if err != nil {
```
sethvargo, line 55:
```suggestion
			if _, err = gh.CreateIssueComment(ctx, owner, repo, *issueToClose.Number, "Drift Resolved."); err != nil {
```

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+				return fmt.Errorf("failed to comment on issue %s/%s %d: %w", owner, repo, issueToClose.Number, err)
+			}
+			err = gh.CloseIssue(ctx, owner, repo, *issueToClose.Number)
+			if err != nil {
```
sethvargo, line 59:
```suggestion
			if err := gh.CloseIssue(ctx, owner, repo, *issueToClose.Number); err != nil {
```

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+
+func createUpdateOrCloseIssues(ctx context.Context, token, owner, repo string, assignees, labels []string, drift *IAMDrift) error {
+	if len(labels) == 0 {
+		return fmt.Errorf("invalid argument - at least one 'label' must be provided")
+	}
+
+	changesDetected := len(drift.ClickOpsChanges) > 0 || len(drift.MissingTerraformChanges) > 0
+	gh := github.NewClient(ctx, token)
+
+	issues, err := gh.ListIssues(ctx, owner, repo, labels, github.Open)
```
sethvargo, line 47: Does this actually list all issues? I thought GitHub used pagination in their APIs.

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create GitHub for %s/%s: %w", owner, repo, err)
+		}
+		issueNumber = *issue.Number
+	} else {
+		issueNumber = *issues[0].Number
+	}
+
+	_, err = gh.CreateIssueComment(ctx, owner, repo, issueNumber, driftMessage(drift))
+	if err != nil {
```
sethvargo, line 78:
```suggestion
	if _, err = gh.CreateIssueComment(ctx, owner, repo, issueNumber, driftMessage(drift)); err != nil {
```

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("failed to comment on issue %s/%s %d: %w", owner, repo, issueNumber, err)
+	}
+
+	return nil
+}
+
+func driftMessage(drift *IAMDrift) string {
+	var msg string
```
sethvargo, line 86: Use `strings.Builder` or `bytes.Buffer` for string concatenation instead.

pkg/github/github.go
```diff
@@ -93,13 +117,102 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+// ListIssues lists all issues and returns their numbers in a repository matching the given criteria.
+func (g *GitHubClient) ListIssues(ctx context.Context, owner, repo string, labels []string, state string) ([]*Issue, error) {
+	backoff := retry.NewFibonacci(g.cfg.initialRetryDelay)
+	backoff = retry.WithMaxRetries(g.cfg.maxRetries, backoff)
+	backoff = retry.WithCappedDuration(g.cfg.maxRetryDelay, backoff)
+
+	var response []*Issue
+
+	if err := retry.Do(ctx, backoff, func(ctx context.Context) error {
+		issues, resp, err := g.client.Issues.ListByRepo(ctx, owner, repo, &github.IssueListByRepoOptions{
```
sethvargo, line 129: I think this can paginate

pkg/github/github.go
```diff
@@ -93,13 +117,102 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+			if _, ok := ignoredStatusCodes[resp.StatusCode]; !ok {
+				return retry.RetryableError(err)
+			}
+
+			return fmt.Errorf("failed to list issues: %w", err)
+		}
+
+		for _, i := range issues {
+			response = append(response, &Issue{Number: i.Number})
+		}
```
sethvargo, line 143: There's a race condition here with the retry logic. It's possible that new issues have been created while you're paginating, in which case you'd likely end up with duplicates. It would be better to accumulate all these issues into a map keyed by a unique ID (e.g. issue number) and then convert to a slice once you're done. You could also just return the map.

pkg/github/github.go
```diff
@@ -93,13 +117,102 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list issues after retries: %w", err)
+	}
+
+	return response, nil
+}
+
+// CreateIssue creates an issue.
+func (g *GitHubClient) CreateIssue(ctx context.Context, owner, repo, title, body string, assignees, labels []string) (*Issue, error) {
+	backoff := retry.NewFibonacci(g.cfg.initialRetryDelay)
```
sethvargo, line 154: I think this entire block (sans the response) everywhere could be encapsulated in a helper function like `g.withRetries()` to make things cleaner.

pkg/github/github.go
```diff
@@ -40,10 +40,34 @@ type Config struct {
 	maxRetryDelay     time.Duration
 }
 
+// Issue is the GitHub Issue.
+type Issue struct {
+	Number *int
```
sethvargo, line 45: Why make these pointers? Do we expect nil ints?

pkg/github/github.go
```diff
@@ -40,10 +40,34 @@ type Config struct {
...
+}
+
+type IssueComment struct {
+	ID *int64
```
sethvargo, line 49: Why make these pointers? Do we expect nil ints?

pkg/github/github.go
```diff
@@ -40,10 +40,34 @@ type Config struct {
...
+}
+
+var (
+	Closed = "closed"
+	Open   = "open"
+	Any    = "any"
```
sethvargo, line 55: Doc. Also recommend `const` instead of `var`.

sethvargo (changes requested):

pkg/github/github.go
dcreey: Updated.

pkg/github/github.go
dcreey: Done

pkg/github/github.go
dcreey: I used var so that I could reference the ptr since we need a ptr in the api.

pkg/github/github.go
dcreey: Done

pkg/github/github.go
dcreey: Done

pkg/commands/drift/github.go
dcreey: Done

pkg/commands/drift/github.go
dcreey: Done

drevell (commented):
Withdrawing myself from this review because Seth and Mike are already here, I don't think we need a third reviewer

pkg/commands/drift/github.go
dcreey: Added a comment. Basically we want to always filter on the label when we list issues - that way we only ever have to consider "drift" issues.

pkg/commands/drift/github.go
dcreey: Broke it out into two methods.

pkg/commands/drift/github.go
dcreey: Done.

pkg/commands/drift/command.go
dcreey: Refactored some.

pkg/commands/drift/command.go
```diff
@@ -122,20 +166,23 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
...
-		uris := keys(iamDiff.MissingTerraformChanges)
-		sort.Strings(uris)
-		c.Outf("Found Missing Terraform Changes \n> %s", strings.Join(uris, "\n> "))
+
+	if c.flagSkipGitHubIssue {
+		return nil
+	}
+	if changesDetected {
+		if err = createOrUpdateIssue(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueAssignees, c.flagGitHubIssueLabels, m); err != nil {
+			return fmt.Errorf("failed to manage GitHub Issue %w", err)
```
sethvargo, line 180:
```suggestion
			return fmt.Errorf("failed to manage GitHub Issue: %w", err)
```

pkg/github/github.go
```diff
@@ -93,13 +117,118 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
 	return g
 }
 
+// ListIssues lists all issues and returns their numbers in a repository matching the given criteria.
+func (g *GitHubClient) ListIssues(ctx context.Context, owner, repo string, labels []string, state string) ([]*Issue, error) {
```
sethvargo, line 121: This is a generic function, so it would probably be better to accept a `github.issueListByRepoOptions` instead of labels. It's not immediately clear that this filters by labels (nor is it reusable for other queries).

dcreey: Done

pkg/commands/drift/command.go
```diff
@@ -122,20 +166,23 @@ func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
...
+		}
+	} else {
+		if err = closeIssues(ctx, c.flagGitHubToken, c.flagGitHubOwner, c.flagGitHubRepo, c.flagGitHubIssueLabels); err != nil {
+			return fmt.Errorf("failed to close GitHub Issues %w", err)
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/command.go
dcreey: Done

pkg/github/github.go
dcreey: Done

pkg/commands/drift/command.go
```diff
@@ -85,6 +89,46 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+	f.BoolVar(&cli.BoolVar{
+		Name:    "skip-github-issue",
+		Target:  &c.flagSkipGitHubIssue,
+		Example: "true",
+		Usage:   `Whether or not to create a GitHub Issue when a drift is detected.`,
+		Default: false,
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "github-token",
+		Target:  &c.flagGitHubToken,
```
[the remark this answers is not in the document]
dcreey: Done

pkg/github/github.go
```diff
@@ -93,15 +117,111 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+			return nil, fmt.Errorf("failed to list issues after retries: %w", err)
+		}
+	}
+
+	response := make([]*Issue, len(uniqueResponses))
+	i := 0
+	for _, r := range uniqueResponses {
+		response[i] = r
+		i++
+	}
```
sethvargo, line 158: It would be more idiomatic to do:

```suggestion
	response := make([]*Issue, 0, len(uniqueResponses))
	for _, r := range uniqueResponses {
		responses = append(responses, r)
	}
```

pkg/github/github.go
```diff
@@ -188,13 +294,25 @@ func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string
...
 	}); err != nil {
-		return nil, nil, fmt.Errorf("failed to list pull request comments: %w", err)
+		return nil, fmt.Errorf("failed to list pull request comments: %w", err)
 	}
 
-	return commentsResponse, gitHubResponse, nil
+	return commentsResponse, nil
+}
+
+func (g *GitHubClient) withRetries(ctx context.Context, retryFunc func(ctx context.Context) error) error {
```
sethvargo, line 309: Nit:

```suggestion
func (g *GitHubClient) withRetries(ctx context.Context, retryFunc retry.RetryFunc {
```

pkg/commands/drift/command.go
```diff
@@ -85,6 +89,47 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		EnvVar:  "GITHUB_TOKEN",
+		Default: "",
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/command.go
```diff
@@ -85,6 +89,47 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+		Target:  &c.flagGitHubOwner,
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
+	})
+	f.StringVar(&cli.StringVar{
+		Name:    "github-repo",
+		Target:  &c.flagGitHubRepo,
+		Example: "guardian",
+		Usage:   `The github token to use to authenticate to create & manage GitHub Issues.`,
+		Default: "",
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/command.go
```diff
@@ -85,6 +89,47 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
...
+	})
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "github-issue-assignees",
+		Target:  &c.flagGitHubIssueAssignees,
+		Example: "dcreey",
+		Usage:   `The assignees to assign to for any created GitHub Issues.`,
+		Default: []string{},
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func createOrUpdateIssue(ctx context.Context, token, owner, repo string, assignees, labels []string, message string) error {
+	// Labels are used to uniquely identify Drift issues.
+	if len(labels) == 0 {
+		return fmt.Errorf("invalid argument - at least one 'label' must be provided")
+	}
+	gh := github.NewClient(ctx, token)
+
+	issues, err := gh.ListIssues(ctx, owner, repo, &githubAPI.IssueListByRepoOptions{Labels: labels, State: github.Open})
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+		issue, err := gh.CreateIssue(ctx, owner, repo, issueTitle, issueBody, assignees, labels)
+		if err != nil {
+			return fmt.Errorf("failed to create GitHub issue for %s/%s with assignees %s and labels %s: %w", owner, repo, assignees, labels, err)
+		}
+		issueNumber = issue.Number
+	} else {
+		issueNumber = issues[0].Number
+	}
+
+	if _, err = gh.CreateIssueComment(ctx, owner, repo, issueNumber, message); err != nil {
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func closeIssues(ctx context.Context, token, owner, repo string, labels []string) error {
+	// Labels are used to uniquely identify Drift issues.
+	if len(labels) == 0 {
+		return fmt.Errorf("invalid argument - at least one 'label' must be provided")
+	}
+	gh := github.NewClient(ctx, token)
+	issues, err := gh.ListIssues(ctx, owner, repo, &githubAPI.IssueListByRepoOptions{Labels: labels, State: github.Open})
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to list GitHub issues for %s/%s: %w", owner, repo, err)
+	}
+	for _, issueToClose := range issues {
+		if _, err = gh.CreateIssueComment(ctx, owner, repo, issueToClose.Number, "Drift Resolved."); err != nil {
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+			return fmt.Errorf("failed to comment on issue %s/%s %d: %w", owner, repo, issueToClose.Number, err)
+		}
+		if err = gh.CloseIssue(ctx, owner, repo, issueToClose.Number); err != nil {
```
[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/github.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+	return nil
+}
+
+func driftMessage(drift *IAMDrift) string {
+	var msg strings.Builder
+	if len(drift.ClickOpsChanges) > 0 {
+		uris := keys(drift.ClickOpsChanges)
+		sort.Strings(uris)
```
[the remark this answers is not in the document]
dcreey: Done

pkg/github/github.go
dcreey: Made them constants. Found a way. We use these outside of this module so I wanted to keep them public.

pkg/github/github.go
dcreey: Done. This is what I wanted but the linter complained without the `0` arg.

pkg/github/github.go
dcreey: Done

sethvargo (approved):

PR #135, opened 2023-07-14
fix(drift): Context cancelled issue

dcreey:
Fails to download subsequent TF state files because the ctx::Cancel() was already called.

### DIFF
--- a/pkg/storage/google_cloud_storage.go
+++ b/pkg/storage/google_cloud_storage.go
@@ -154,16 +154,16 @@ func (s *GoogleCloudStorage) UploadObject(ctx context.Context, bucket, name stri
 }
 
 // DownloadObject downloads an object from a Google Cloud Storage bucket. The caller must call Close on the returned Reader when done reading.
-func (s *GoogleCloudStorage) DownloadObject(ctx context.Context, bucket, name string) (io.ReadCloser, error) {
-	o, ctx, _ := s.objectHandleWithRetries(ctx, bucket, name)
-	// defer cancel()
+func (s *GoogleCloudStorage) DownloadObject(ctx context.Context, bucket, name string) (io.ReadCloser, context.CancelFunc, error) {
+	o, ctx, cancel := s.objectHandleWithRetries(ctx, bucket, name)
 
 	r, err := o.NewReader(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get google cloud storage reader: %w", err)
+		cancel()
+		return nil, nil, fmt.Errorf("failed to get google cloud storage reader: %w", err)
 	}
 
-	return r, nil
+	return r, cancel, nil
 }
 
 // ObjectMetadata gets the metadata for a Google Cloud Storage object.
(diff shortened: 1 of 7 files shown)


pkg/terraform/parser/parser_test.go
```diff
@@ -209,7 +213,11 @@ func TestParser_ProcessStates(t *testing.T) {
 				t.Errorf("ProcessStates() failed to read json file %v", err)
 			}
 
-			gcsClient := &storage.MockStorageClient{GetData: string(data), GetErr: tc.wantErr}
+			gcsClient := &storage.MockStorageClient{
+				GetData:       string(data),
+				GetErr:        tc.wantErr,
+				GetCancelFunc: func() {},
```
verbanicm, line 219: nit: can we rename these `Get*` -> `Download*`

dcreey: Done

verbanicm (approved):

yolocs (approved):

drevell (approved):

pkg/storage/google_cloud_storage.go
```diff
@@ -154,16 +154,16 @@ func (s *GoogleCloudStorage) UploadObject(ctx context.Context, bucket, name stri
 }
 
 // DownloadObject downloads an object from a Google Cloud Storage bucket. The caller must call Close on the returned Reader when done reading.
-func (s *GoogleCloudStorage) DownloadObject(ctx context.Context, bucket, name string) (io.ReadCloser, error) {
+func (s *GoogleCloudStorage) DownloadObject(ctx context.Context, bucket, name string) (io.ReadCloser, context.CancelFunc, error) {
 	o, ctx, cancel := s.objectHandleWithRetries(ctx, bucket, name)
-	defer cancel()
```
sethvargo, line 159: This is a little bit weird, because now we're leaking our internal abstraction outwards.

The user is providing a context in the call signature, and it's up to them to manage that context (that's the rule of contexts). We are creating our own internal context for the purposes of forcing a timeout on the retries.

We want to ensure the retryContext is cancelled when the user is done reading, since that frees up the context's resources. Therefore I think we want this:

```go
type readCloserCanceller struct {
  rc io.ReadCloser
  cancelFunc context.CancelFunc
}

func (r *readCloserCanceller) Close() error {
  defer r.cancelFunc()
  return r.Close()
}

func DownloadObject() {
  r, err := o.NewReader(ctx)
  if err != nil {
    cancel()
    return nil, fmt.Errorf("failed to get google cloud storage reader: %w", err)
  }

  return &readCloserCanceller{
    rc: r,
    cancelFunc: cancel,
  }, nil
}
```

sethvargo: https://github.com/abcxyz/guardian/pull/138

dcreey: Ack. The other thing I considered was fully encapsulating by accepting an argument that is a marshalling func like you did in the original impl for the json unmarshalling of tf state. Either would be better. I'll make an issue to address this. https://github.com/abcxyz/guardian/issues/137

PR #136, opened 2023-07-14
feat: guardian plan

verbanicm:
- Decided to use buffer instead of file and reset it after each run. The output written to the buffer should be minimal.

### DIFF
--- a/.github/workflows/test-plan.yml
+++ b/.github/workflows/test-plan.yml
@@ -58,7 +58,7 @@ jobs:
           strategy: 'delete'
 
   plan:
-    if: '${{ needs.init.outputs.directories != ''[]'' }}'
+    if: "${{ needs.init.outputs.directories != '[]' }}"
     needs:
       - 'init'
     runs-on: 'ubuntu-latest'
@@ -110,16 +110,15 @@ jobs:
           git config --global url."https://guardian-bot:${{ steps.mint-token.outputs.token }}@github.com".insteadOf "https://github.com"
 
       - name: 'Guardian Plan'
-        shell: 'bash'
         env:
           GITHUB_TOKEN: '${{ github.token }}'
           WORKING_DIRECTORY: './${{ matrix.working_directory }}'
           BUCKET_NAME: '${{ vars.GUARDIAN_BUCKET_NAME }}'
         run: |-
           guardian plan \
-            --github-token=$GITHUB_TOKEN \
-            --working-directory=$WORKING_DIRECTORY \
-            --bucket-name=$BUCKET_NAME
+            --github-token=${GITHUB_TOKEN} \
+            --working-directory=${WORKING_DIRECTORY} \
+            --bucket-name=${BUCKET_NAME}
 
   plan_success:
     if: '${{ always() }}'
--- a/.github/workflows/test-plan.yml
+++ b/.github/workflows/test-plan.yml
@@ -72,6 +72,23 @@ jobs:
         with:
           ref: '${{ github.event.pull_request.head.ref }}'
 
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568' # ratchet:actions/setup-go@v3
+        with:
+          go-version: '1.20'
+
+      - name: 'Install Guardian CLI'
+        shell: 'bash'
+        run: 'go install github.com/abcxyz/guardian/cmd/guardian@${{ github.event.pull_request.head.sha }}'
+
+      - name: 'Setup Terraform'
+        uses: 'abcxyz/secure-setup-terraform@78e3f2fdfb7ebf59ce42141f6dd93bdd65bf04ed' # ratchet:abcxyz/secure-setup-terraform@v0.2.5
+        with:
+          terraform_version: '1.3.6'
+          terraform_module_location: './${{ matrix.working_directory }}'
+          terraform_lockfile_location: './${{ matrix.working_directory }}'
+          terraform_wrapper: false
+
       - name: 'Mint Token'
         id: 'mint-token'
         uses: 'abcxyz/github-token-minter/.github/actions/mint-token@c6ea9b91e3e1df75a6e86dad8e5b0a0a45b42fec'
@@ -93,13 +110,16 @@ jobs:
           git config --global url."https://guardian-bot:${{ steps.mint-token.outputs.token }}@github.com".insteadOf "https://github.com"
 
       - name: 'Guardian Plan'
-        uses: './.github/actions/plan'
+        shell: 'bash'
         env:
-          GITHUB_TOKEN: '${{ steps.mint-token.outputs.token }}'
-        with:
-          working_directory: '${{ matrix.working_directory }}'
-          bucket_name: '${{ vars.GUARDIAN_BUCKET_NAME }}'
-          terraform_version: '1.3.6'
+          GITHUB_TOKEN: '${{ github.token }}'
+          WORKING_DIRECTORY: './${{ matrix.working_directory }}'
+          BUCKET_NAME: '${{ vars.GUARDIAN_BUCKET_NAME }}'
+        run: |-
+          guardian plan \
+            --github-token=$GITHUB_TOKEN \
+            --working-directory=$WORKING_DIRECTORY \
+            --bucket-name=$BUCKET_NAME
 
   plan_success:
     if: '${{ always() }}'
--- a/pkg/commands/plan/config.go
+++ b/pkg/commands/plan/config.go
@@ -1,0 +1,61 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package plan
+
+import (
+	"fmt"
+
+	"github.com/sethvargo/go-githubactions"
+)
+
+// Config defines the of configuration required for running the plan action.
+type Config struct {
+	// GitHub context values
+	IsAction          bool
+	EventName         string
+	RepositoryOwner   string
+	RepositoryName    string
+	PullRequestNumber int
+	ServerURL         string
+	RunID             int64
+	RunAttempt        int64
+}
+
+// MapGitHubContext maps values from the GitHub context to the PlanConfig.
+func (c *Config) MapGitHubContext(context *githubactions.GitHubContext) error {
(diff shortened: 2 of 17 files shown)


pkg/commands/plan/config.go
```diff
@@ -0,0 +1,61 @@
...
+		return fmt.Errorf("failed to get pull request number from github event")
+	}
+
+	// parsing a json file into this map causes javascript numbers to be float64
+	v, ok := context.Event["number"].(float64)
+	if !ok {
+		return fmt.Errorf("failed to get pull request number, github.event.number is not a number")
+	}
+
+	c.PullRequestNumber = int(v)
```
sethvargo, line 58: It'd be better to use `json.UseNumber()` and then `n.Int64()`.

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,281 @@
...
+		return fmt.Errorf("failed to create start comment: %w", err)
+	}
+
+	c.Outf("\n\nRunning Terraform commands...")
+	result, err := c.handleTerraformPlan(ctx)
+
+	if err != nil {
+		var comment strings.Builder
+
+		comment.WriteString(fmt.Sprintf("%s 🟥 Failed for dir: `%s` %s\n\n<details>\n<summary>Error</summary>\n\n```\n\n%s\n```\n</details>", planCommentPrefix, c.flagWorkingDirectory, logURL, err))
```
sethvargo, line 72: Nit, it's more efficienter:

```suggestion
		fmt.Fprintf(&comment, "%s 🟥 Failed for dir: `%s` %s\n\n<details>\n<summary>Error</summary>\n\n```\n\n%s\n```\n</details>", planCommentPrefix, c.flagWorkingDirectory, logURL, err)
```

sethvargo: (Throughout)

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,281 @@
...
+		}
+
+		if _, commentErr := c.githubClient.CreateIssueComment(
+			ctx,
+			c.cfg.RepositoryOwner,
+			c.cfg.RepositoryName,
+			c.cfg.PullRequestNumber,
+			comment.String(),
+		); commentErr != nil {
+			merr = errors.Join(merr, fmt.Errorf("failed to create plan error comment: %w", commentErr))
```
sethvargo, line 84: Do you want to return after this error?

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,281 @@
...
+		Input:            util.Ptr[bool](false),
+		NoColor:          util.Ptr[bool](true),
+		DetailedExitcode: util.Ptr[bool](true),
+		LockTimeout:      util.Ptr[string](c.flagLockTimeout.String()),
+	})
+
+	stderr.Reset()
+	c.actions.EndGroup()
+
+	hasChanges := planExitCode == 2
```
sethvargo, line 189: Document?

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,281 @@
...
+		return &Result{commentDetails: stderr.String()}, fmt.Errorf("failed to plan: %w", err)
+	}
+
+	if !hasChanges {
+		return &Result{hasChanges: hasChanges}, nil
+	}
+
+	c.actions.Group("\n\nFormatting plan output")
+
+	_, err = c.terraformClient.Show(ctx, multiStdout, multiStderr, &terraform.ShowOptions{File: util.Ptr[string](c.planFilename), NoColor: util.Ptr[bool](true)})
```
sethvargo, line 201: Do you really need to box these generics? `util.Ptr` should know what it's giving/getting...

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,281 @@
...
-func (p *PlanCommand) Run(ctx context.Context, args []string) error {
-	f := p.Flags()
-	if err := f.Parse(args); err != nil {
-		return fmt.Errorf("failed to parse flags: %w", err)
+// handleUploadGuardianPlan uploads the Guardian plan binary to the configured Guardian storage bucket.
+func (c *PlanCommand) handleUploadGuardianPlan(ctx context.Context, planFilePath string, planData []byte, planExitCode int) error {
+	guardianPlanName := fmt.Sprintf("guardian-plans/%s/%s/%d/%s", c.cfg.RepositoryOwner, c.cfg.RepositoryName, c.cfg.PullRequestNumber, planFilePath)
+
+	metadata := make(map[string]string)
+	metadata["plan_exit_code"] = fmt.Sprintf("%d", planExitCode)
```
sethvargo, line 244:
```suggestion
	metadata["plan_exit_code"] = strconv.FormatInt(planExitCode, 10)
```

pkg/commands/plan/plan.go
verbanicm: Wasnt sure, I will remove.

pkg/commands/plan/config.go
verbanicm: This didnt work:

```panic: interface conversion: interface {} is float64, not json.Number [recovered]
        panic: interface conversion: interface {} is float64, not json.Number```

File: pkg/commands/plan/plan.go (hunk 6 above)
Comment (verbanicm) [reply pkg/commands/plan/plan.go:244]: Needs to be int not int64, used strconv.Itoa()

File: pkg/commands/plan/plan.go (hunk 3 above)
Comment (verbanicm) [reply pkg/commands/plan/plan.go:84]: Hmmm, this needs to be refactored because of multi-dir to single dir. Will update.

File: .github/workflows/test-plan.yml
Comment (verbanicm) [inline .github/workflows/test-plan.yml:1]: This removes the need for an action and embeds the CLI use into the workflow.

This seems like a cleaner API by having users pass values directly to the CLI command rather than passing through the action inputs.

This also decouples Guardian from secure-setup-terraform, which I dont know why but allows people to configure Terraform however they want. We will still recommend and showcase the use of secure-setup-terraform in our examples.

Comment (verbanicm) [inline .github/workflows/test-plan.yml:1]: This removes the need for an action and embeds the CLI use into the workflow.

This seems like a cleaner API by having users pass values directly to the CLI command rather than passing through the action inputs.

This also decouples Guardian from secure-setup-terraform, which I dont know why but allows people to configure Terraform however they want. We will still recommend and showcase the use of secure-setup-terraform in our examples.

File: .github/workflows/test-plan.yml [hunk 7]
```diff
@@ -72,6 +72,23 @@ jobs:
...
+          go-version: '1.20'
+
+      - name: 'Install Guardian CLI'
+        shell: 'bash'
+        run: 'go install github.com/abcxyz/guardian/cmd/guardian@${{ github.event.pull_request.head.sha }}'
+
+      - name: 'Setup Terraform'
+        uses: 'abcxyz/secure-setup-terraform@78e3f2fdfb7ebf59ce42141f6dd93bdd65bf04ed' # ratchet:abcxyz/secure-setup-terraform@v0.2.5
+        with:
+          terraform_version: '1.3.6'
```
sethvargo, line 87: Should this be configurable (can be a follow-up)?

pkg/commands/plan/config.go
sethvargo: Then you should use `int(math.Round(v))`

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,241 @@
...
-func (p *PlanCommand) Run(ctx context.Context, args []string) error {
-	f := p.Flags()
-	if err := f.Parse(args); err != nil {
-		return fmt.Errorf("failed to parse flags: %w", err)
+// handleTerraformPlan runs the required Terraform commands for a full run of
+// a Guardian plan using the Terraform CLI.
+func (c *PlanCommand) handleTerraformPlan(ctx context.Context) (*Result, error) {
+	var stdout, stderr strings.Builder
+	multiStdout := io.MultiWriter(c.Stdout(), &stdout)
+	multiStderr := io.MultiWriter(c.Stderr(), &stderr)
```
sethvargo, line 125: Nice

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,241 @@
...
 	}
 
-	args = f.Args()
-	if len(args) > 0 {
-		return fmt.Errorf("unexpected arguments: %q", args)
+	c.actions.Group("Initializing Terraform")
+
+	if _, err := c.terraformClient.Init(
+		ctx,
+		c.Stdout(),
```
sethvargo, line 136: Should this be `multiStdout`?

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,241 @@
...
+		return &Result{commentDetails: stderr.String()}, fmt.Errorf("failed to initialize: %w", err)
+	}
+	stderr.Reset()
+	c.actions.EndGroup()
+
+	c.actions.Group("Validating Terraform")
+
+	if _, err := c.terraformClient.Validate(
+		ctx,
+		c.Stdout(),
```
sethvargo, line 153: Should this be `multiStdout`?

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,241 @@
...
+		return &Result{commentDetails: stderr.String()}, fmt.Errorf("failed to validate: %w", err)
 	}
 
-	p.Outf("Running Guardian plan...")
+	stderr.Reset()
+	c.actions.EndGroup()
+
+	c.actions.Group("Running Terraform plan")
+
+	planExitCode, err := c.terraformClient.Plan(ctx, c.Stdout(), multiStderr, &terraform.PlanOptions{
```
sethvargo, line 165: Should this be `multiStdout`?

pkg/commands/plan/plan.go
```diff
@@ -4,61 +4,241 @@
...
+}
+
+// handleUploadGuardianPlan uploads the Guardian plan binary to the configured Guardian storage bucket.
+func (c *PlanCommand) handleUploadGuardianPlan(ctx context.Context, planFilePath string, planData []byte, planExitCode int) error {
+	guardianPlanName := fmt.Sprintf("guardian-plans/%s/%s/%d/%s", c.cfg.RepositoryOwner, c.cfg.RepositoryName, c.cfg.PullRequestNumber, planFilePath)
+
+	metadata := make(map[string]string)
+	metadata["plan_exit_code"] = strconv.Itoa(planExitCode)
+
+	if err := c.storageClient.UploadObject(ctx, c.flagBucketName, guardianPlanName, planData,
```
sethvargo, line 235: As a follow-up, we should probably be gzipping/compressing these objects

pkg/commands/plan/plan.go
verbanicm: Ditto

pkg/commands/plan/plan.go
verbanicm: Ditto

.github/workflows/test-plan.yml
verbanicm: Do you mean as a matrix strategy for testing purposes? Otherwise, this is our test workflow we can update as needed.

pkg/commands/plan/plan.go
sethvargo: Ah ok

sethvargo (approved):

PR #142, opened 2023-07-17
feat: add plan init command and refactor

verbanicm:
Implement the plan init command. This step is required to bootstrap a pull request and generate the target directories. This also refactors the exiting plan command to use a subcommand for init and run allowing for future expansion of other commands related to plan.


pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,240 @@
...
+		Target:  &c.flagDirectories,
+		Example: "dir1,dir2,dir3",
+		Usage:   "The subset of directories to detect files changes on.",
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "delete-outdated-comments",
+		Target:  &c.flagDeleteOutdatedComments,
+		Default: true,
+		Usage:   "The working directory for Guardian to execute commands in.",
```
drevell, line 119: This Usage text looks like a copy-paste oversight

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,240 @@
...
+		dirs, err = terraform.GetEntrypointDirectories(c.flagWorkingDirectory)
+		if err != nil {
+			return fmt.Errorf("failed to find terraform directories: %w", err)
+		}
+		logger.Debugw("terraform directories", "dirs", dirs)
+	}
+
+	diffDirs, err := c.gitClient.DiffDirs(ctx, c.flagDestRef, c.flagSourceRef)
+	if err != nil {
+		c.Errf("ERROR: failed to find changed directories")
```
drevell, line 198: Is it conventional to put `ERROR:` in error logs? Seems maybe redundant

pkg/commands/plan/plan_run.go
```diff
@@ -22,36 +22,214 @@ import (
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create google cloud storage client: %w", err)
+	}
+	c.storageClient = sc
+
+	return c.Process(ctx)
+}
+
+func normalizeWorkingDir(dir string) (string, error) {
```
drevell, line 192: Could you drop a comment on this function explaining what "normalize" means in this context?

drevell (changes requested):

drevell (commented):
Recent team convention is to put a semi-detailed writeup as a PR description. This PR seems like a great candidate for that, containing what/why/how

drevell (dismissed):

pkg/github/github.go
```diff
@@ -281,8 +283,9 @@ func (g *GitHubClient) DeleteIssueComment(ctx context.Context, owner, repo strin
 }
 
 // ListIssueComments lists existing comments for an issue or pull request.
-func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*IssueComment, error) {
+func (g *GitHubClient) ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*IssueComment, int, error) {
```
sethvargo, line 286: Why this API change?

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if c.flagJSON {
+		outJSON, err := json.Marshal(targetDirs)
+		if err != nil {
+			return fmt.Errorf("failed to create json string: %w", err)
+		}
+		c.Outf(string(outJSON))
+	} else {
+		for _, dir := range targetDirs {
+			c.Outf("%s", dir)
```
sethvargo, line 215: Do you need a newline after this?

verbanicm: Outf handles that for you....which I am not sure I agree with but for now, it works.

https://github.com/abcxyz/pkg/blob/main/cli/command.go#L302

pkg/github/github.go
verbanicm: Paging, let me add a struct instead with paging details

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+
+	gitClient    git.Git
+	githubClient github.GitHub
+
+	// testFlagSetOpts is only used for testing.
+	testFlagSetOpts []cli.Option
+}
+
+func (c *PlanInitCommand) Desc() string {
+	return `Run the Terraform plan for a directory.`
```
sethvargo, line 61: Nit: usually we don't put a period in `Desc`

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+		Name:    "directories",
+		Target:  &c.flagDirectories,
+		Example: "dir1,dir2,dir3",
+		Usage:   "The subset of directories to target for Terraform planning.",
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "delete-outdated-comments",
+		Target:  &c.flagDeleteOutdatedComments,
+		Default: true,
```
sethvargo, line 111: This is more of a Go-style thing. Usually booleans default to "false" (since the zero value is false), so it'd be more appropriate to have:

```shell
-keep-outdated-comments
```

and default it to false.

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+		ctx,
+		c.GitHubFlags.FlagGitHubToken,
+		github.WithInitialRetryDelay(c.RetryFlags.FlagInitialRetryDelay),
+		github.WithMaxRetries(c.RetryFlags.FlagMaxRetries),
+		github.WithMaxRetryDelay(c.RetryFlags.FlagMaxRetryDelay),
+	)
+
+	if len(c.flagDirectories) == 0 {
+		c.flagDirectories = append(c.flagDirectories, cwd)
+	}
```
sethvargo, line 155: This behavior seems a little weird? If you don't specify any directories, it defaults to the cwd? That should at least be documented.

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+
+	c.entrypoints = dirs
+	logger.Debugf("terraform entrypoint directories: %s", c.entrypoints)
+
+	return c.Process(ctx)
+}
+
+// Process handles the main logic for the Guardian plan init process.
+func (c *PlanInitCommand) Process(ctx context.Context) error {
+	logger := logging.FromContext(ctx)
```
sethvargo, line 172: I'd probably add a bunch of context to the logger here:

```suggestion
	logger := logging.FromContext(ctx).
		Named("plan.process").
		With("github_owner", c.GitHubFlags.FlagGitHubOwner).
		With("github_repo", ...)
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+	logger.Debugln("starting Guardian plan init")
```
sethvargo, line 173:
```suggestion
	logger.Debug("starting Guardian plan init")
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if c.GitHubFlags.FlagGitHubAction && c.flagDeleteOutdatedComments {
+		logger.Debugln("removing outdated comments...")
```
sethvargo, line 176:
```suggestion
		logger.Debug("removing outdated comments...")
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+		if err := c.deleteOutdatedComments(ctx, c.GitHubFlags.FlagGitHubOwner, c.GitHubFlags.FlagGitHubRepo, c.flagPullRequestNumber); err != nil {
+			return fmt.Errorf("failed to delete outdated comments: %w", err)
+		}
+	}
+
+	diffDirs, err := c.gitClient.DiffDirsAbs(ctx, c.flagDestRef, c.flagSourceRef)
+	if err != nil {
+		return fmt.Errorf("failed to find git diff directories: %w", err)
+	}
+	logger.Debugf("changed directories: %+v", diffDirs)
```
sethvargo, line 186:
```suggestion
	logger.Debugw("computed directories", "directories", diffDirs)
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+
+	targetDirs := util.GetSliceIntersection(c.entrypoints, diffDirs)
+	logger.Debugf("target directories: %+v", targetDirs)
```
sethvargo, line 189:
```suggestion
	logger.Debugw("target directories", "target_directories", targetDirs)
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+	// creates an ugly github workflow name
+	for k, dir := range targetDirs {
+		childPath, err := util.ChildPath(c.workingDir, dir)
+		if err != nil {
+			return fmt.Errorf("failed to get child path for: %w", err)
+		}
+		targetDirs[k] = childPath
+	}
+
+	if c.flagJSON {
```
sethvargo, line 201: Per above, I'm not sure this is the best approach for formatting.

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+		outJSON, err := json.Marshal(targetDirs)
+		if err != nil {
+			return fmt.Errorf("failed to create json string: %w", err)
+		}
+		c.Outf(string(outJSON))
```
sethvargo, line 206:
```suggestion
		if err := json.NewEncoder(c.Stdout()).Encode(targetDirs); err != nil {
			return fmt.Errorf("failed to create json string: %w", err)
		}
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to list comments: %w", err)
+		}
+
+		for _, comment := range comments {
+			if strings.HasPrefix(comment.Body, planCommentPrefix) {
+				if err := c.githubClient.DeleteIssueComment(ctx, owner, repo, comment.ID); err != nil {
+					return fmt.Errorf("failed to delete comment: %w", err)
+				}
+			}
```
sethvargo, line 233: Style nit, usually we try to avoid this level of deep nesting:

```suggestion
			if !strings.HasPrefix(comment.Body, planCommentPrefix) {
				continue
			}

			if err := c.githubClient.DeleteIssueComment(ctx, owner, repo, comment.ID); err != nil {
				return fmt.Errorf("failed to delete comment: %w", err)
			}
```

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// deleteOutdatedComments deletes the pull request comments from previous Guardian plan runs.
+func (c *PlanInitCommand) deleteOutdatedComments(ctx context.Context, owner, repo string, number int) error {
+	listOpts := &gh.IssueListCommentsOptions{
+		ListOptions: gh.ListOptions{PerPage: 100},
+	}
+
+	for {
+		comments, paging, err := c.githubClient.ListIssueComments(ctx, owner, repo, number, listOpts)
```
sethvargo, line 223: I see what this changed. From an API perspective, I would expect this to return a `(struct{}, error)` where the structure had two fields:

```go
type ListIssueCommentsResponse struct {
  Comments []*ghapi.Comment
  Pagination *ghapi.Pagination
}
```

pkg/commands/plan/plan_run.go
```diff
@@ -24,35 +24,195 @@ import (
...
+	githubClient    github.GitHub
+	storageClient   storage.Storage
+	terraformClient terraform.Terraform
+
+	// testFlagSetOpts is only used for testing.
+	testFlagSetOpts []cli.Option
+}
+
+func (c *PlanRunCommand) Desc() string {
+	return `Run the Terraform plan for a directory.`
```
sethvargo, line 76:
```suggestion
	return `Run the Terraform plan for a directory`
```

pkg/commands/plan/plan_run.go
```diff
@@ -24,35 +24,195 @@ import (
...
+		Target:  &c.flagBucketName,
+		Example: "my-guardian-state-bucket",
+		Usage:   "The Google Cloud Storage bucket name to store Guardian plan files.",
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "protect-lockfile",
+		Target:  &c.flagProtectLockfile,
+		Default: true,
+		Example: "true",
```
sethvargo, line 120: Same comment about booleans defaulting to false. This should be `-override-lockfile` and default to false.

pkg/commands/plan/testdata/test-tfplan.binary
```diff
@@ -0,0 +1 @@
+this is a plan binary
```
sethvargo, line 1: Is it?

pkg/flags/github.go
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+package flags
+
+import "github.com/abcxyz/pkg/cli"
+
+// GitHubFlags represent the shared GitHub flags among all commands.
+// Embed this struct into any commands that interact with GitHub.
+type GitHubFlags struct {
+	FlagGitHubAction bool
```
sethvargo, line 22: I wonder if this should be `IsGitHubActions` or something. I was confused by the name initially.

pkg/flags/github.go
```diff
@@ -0,0 +1,59 @@
...
+		Target:  &g.FlagGitHubAction,
+		Default: false,
+		Usage:   "Is this running as a GitHub action.",
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:   "github-token",
+		EnvVar: "GITHUB_TOKEN",
+		Target: &g.FlagGitHubToken,
+		Usage:  "The GitHub access token to make GitHub API calls.",
```
sethvargo, line 43: Maybe add "This value is automatically set on GitHub Actions".

pkg/flags/retry.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (r *RetryFlags) AddFlags(set *cli.FlagSet) {
+	f := set.NewSection("Retry options")
+
+	f.Uint64Var(&cli.Uint64Var{
+		Name:    "max-retries",
+		Target:  &r.FlagMaxRetries,
+		Default: uint64(3),
+		Example: "3",
```
sethvargo, line 38: Nit: it's usually better to provide an example that is different from the default (throughout). It provides additional context to the user.

pkg/flags/retry.go
```diff
@@ -0,0 +1,57 @@
...
+	FlagMaxRetries        uint64
+	FlagInitialRetryDelay time.Duration
+	FlagMaxRetryDelay     time.Duration
+}
+
+func (r *RetryFlags) AddFlags(set *cli.FlagSet) {
+	f := set.NewSection("Retry options")
+
+	f.Uint64Var(&cli.Uint64Var{
+		Name:    "max-retries",
```
sethvargo, line 35: I'd suggest we "namespace" these flags and prefix them all with "retry-". We implicitly did this with GitHub, but it would be good to do here too, just so we don't conflict in the future.

pkg/github/github.go
```diff
@@ -77,7 +84,7 @@ type GitHub interface {
 	DeleteIssueComment(ctx context.Context, owner, repo string, id int64) error
 
 	// ListIssueComments lists existing comments for an issue or pull request.
-	ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*IssueComment, error)
+	ListIssueComments(ctx context.Context, owner, repo string, number int, opts *github.IssueListCommentsOptions) ([]*IssueComment, *Paging, error)
```
sethvargo, line 87: Per above, I'd change this return type to be a single struct that has sub-fields "Comments" and "Paging".

pkg/util/util_test.go
```diff
@@ -70,3 +71,51 @@ func TestGetSliceIntersection(t *testing.T) {
...
+			base:   "",
+			target: "",
+			exp:    "",
+		},
+		{
+			name:   "not_child_dir",
+			base:   "./terraform/project",
+			target: ".",
+			err:    "is not a child of",
+		},
```
sethvargo, line 102: Can you add some more exhaustive test cases here, including things like filepaths with spaces and special characters?

sethvargo (changes requested):

.github/workflows/test-plan.yml
```diff
@@ -33,32 +33,53 @@ permissions:
...
     runs-on: 'ubuntu-latest'
     outputs:
-      directories: '${{ steps.dirs.outputs.directories }}'
+      directories: '${{ steps.init.outputs.directories }}'
     steps:
       - name: 'Checkout'
         uses: 'actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c' # ratchet:actions/checkout@v3
         with:
           fetch-depth: 0 # get everything so we can use git diff
-          ref: '${{ github.event.pull_request.head.ref }}'
```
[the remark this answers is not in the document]
verbanicm: Since this is pull_request and not pull_request_target, the head ref is already checked out.

.github/workflows/test-plan.yml
```diff
@@ -33,32 +33,53 @@ permissions:
...
+
+      - name: 'Guardian Plan Init'
+        id: 'init'
+        env:
+          GITHUB_TOKEN: '${{ github.token }}'
+          HEAD_REF: 'origin/${{ github.event.pull_request.head.ref }}'
+          BASE_REF: 'origin/${{ github.event.pull_request.base.ref }}'
+        run: |-
+          DIRS=$(guardian plan init \
+            -directories=terraform \
```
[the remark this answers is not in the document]
verbanicm: This is a value configured by the person configuring guardian. I can add it as a top level env var to make it easier to reason about?

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+var _ cli.Command = (*PlanInitCommand)(nil)
+
+type PlanInitCommand struct {
+	cli.BaseCommand
+
+	workingDir  string
+	entrypoints []string
+
+	flags.GitHubFlags
+	flags.RetryFlags
```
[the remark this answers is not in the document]
verbanicm: +1, seems good so far!

pkg/commands/plan/plan_init.go
```diff
@@ -0,0 +1,243 @@
...
+
+	flagPullRequestNumber      int
```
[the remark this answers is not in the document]
verbanicm: Pull request number is an INT in the call to the github apis, so I matched it.

https://github.com/abcxyz/guardian/blob/main/pkg/github/github.go#L288

pkg/commands/plan/config.go
```diff
@@ -15,48 +15,21 @@
...
 	// GitHub context values
-	IsAction          bool
-	EventName         string
-	RepositoryOwner   string
-	RepositoryName    string
-	PullRequestNumber int
-	ServerURL         string
-	RunID             int64
-	RunAttempt        int64
+	ServerURL  string
```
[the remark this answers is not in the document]
verbanicm: I read this, but I dont understand what action you would like me to take.

.github/workflows/test-plan.yml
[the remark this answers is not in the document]
sethvargo: Yea I think so

pkg/commands/plan/config.go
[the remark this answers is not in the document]
sethvargo: Sorry - I'm saying it's really unclear what this value should be. I see the tests using `https://github.com`, but the value is passed to the `ListCommentsForIssue` function which should be using api.github.com. I think I'm asking you to document this :)

pkg/commands/plan/plan_run.go
verbanicm: Went with --allow-lockfile-changes

pkg/commands/plan/testdata/test-tfplan.binary
verbanicm: lol

pkg/commands/plan/config.go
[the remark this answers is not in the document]
verbanicm: Hmmm, ack. This is only used to generate URLs for comments. The GitHub client doesn't accept and API url override (which I need to add an issue for).

pkg/commands/plan/plan_init.go
verbanicm: will adjust to using args

sethvargo (approved):

PR #149, opened 2023-07-19
fix: always upload plan regardless of changes for apply to read

verbanicm:
Apply reads the plan file and metadata to determine if it should apply, we should always upload the plan file even if no changes.


sethvargo (approved):

verbanicm:
**`🔱 Guardian 🔱 PLAN`** - 🟨 Running for dir: `terraform` [[logs](https://github.com/abcxyz/guardian/actions/runs/1/attempts/1)]

PR #158, opened 2023-07-21
Get flagset from the CLI

sethvargo:
This will inherit the new features in the CLI onto the flag set, such as overriding the environment.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	cloud.google.com/go/asset v1.14.1
 	cloud.google.com/go/storage v1.31.0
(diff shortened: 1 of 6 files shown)


dcreey (dismissed):

sethvargo:
Hmm - I'm seeing these same test failures when I run HEAD locally. Did we break something? @dcreey @verbanicm

verbanicm (approved):

verbanicm (approved):

sethvargo:
I guess this is a situation where readability breaks down, huh. I have readability, but we still want another reviewer...

PR #162, opened 2023-07-21
feat: setup-guardian action

verbanicm:


.github/actions/setup-guardian/action.yml
```diff
@@ -0,0 +1,78 @@
...
+        if [ "${RUNNER_OS}" = "Linux" ]; then
+          OS="linux"
+        elif [ "${RUNNER_OS}" = "macOS" ]; then
+          OS="darwmin"
+        else
+          echo "::error ::Unsupport os ${RUNNER_OS}"
+          exit 1
+        fi
+
+        RELEASE_URL="https://github.com/abcxyz/guardian/releases/download/v${GUARDIAN_VERSION}/guardian_${GUARDIAN_VERSION}_darwin_arm64.tar.gz"
```
dcreey, line 56: Did you mean to use the OS and ARCH variables you set above here?

.github/actions/setup-guardian/action.yml
```diff
@@ -0,0 +1,78 @@
...
+
+inputs:
+  guardian_version:
+    description: 'The version of the Guardian CLI to install.'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: 'Cache Guardian'
```
dcreey, line 28: Is using the cache really worth it? You are just downloading and untarballing the binary - its the same amount of work as the cache AFAIK (cache downloads a tarball and untarballs it).

.github/actions/setup-guardian/action.yml
verbanicm: fixed

.github/actions/setup-guardian/action.yml
verbanicm: This is a common practice for actions, e.g. setup-go, setup-node, etc. I believe you are correct in the process of how the cache works, but my assumption was there less latency or something since its "local to GitHub" (I might be just wishing here).

Also, once Guardian is cached on the main branch it can then be used on feature branches in pull requests and for applies. Since we require guardian two times for plan and apply (to init), I think this makes most sense.

dcreey: Ack. This makes sense in theory and I see no reason not to do it as either the cache should be more performant (or likely eventually will be).

But I still honestly would be curious to see a test to compare the performance of both.

sethvargo: You're both correct. One thing I'll note: if we cache the download, getting future usage metrics can be more challenging. If we don't cache, each download corresponds to a unique run.

.github/actions/setup-guardian/action.yml
```diff
@@ -0,0 +1,80 @@
...
+          OS="linux"
+        elif [ "${RUNNER_OS}" = "macOS" ]; then
+          OS="darmin"
+        else
+          echo "::error ::Unsupported operating system ${RUNNER_OS}"
+          exit 1
+        fi
+
+        RELEASE_URL="https://github.com/abcxyz/guardian/releases/download/v${GUARDIAN_VERSION}/guardian_${GUARDIAN_VERSION}_${OS}_${ARCH}.tar.gz"
+        echo "RELEASE_URL=${RELEASE_URL}" >> $GITHUB_ENV
```
sethvargo, line 59: I wonder if we should generalize this into an action that installs/caches binaries from source? Seems like a common pattern. Then we could also add security features (like checksum verification and signature verification from SLSA).

verbanicm: I like that idea, let me move this to pkg and name it `setup-binary`?

verbanicm:
Moving to abcxyz/pkg as generic `setup-binary` action.

PR #165, opened 2023-07-24
fix: refactor init command to top-level

verbanicm:


sethvargo:
What's the goal with this change? This will prevent us from ever having a guardian-specific initialization separate from Terraform.

verbanicm:
Theres enough of an overlap, that it doesnt make sense to have guardian plan init, guardian apply init, guardian admin init. The only caveat is that plan init should delete plan comments. 

Maybe init is just the wrong word....workflow-init, get-dirs?

PR #166, opened 2023-07-25
feat: Create drift subcommand for determining statefile drift

dcreey:
Statefile drift occurs when the remote GCS bucket contains statefiles not described by local terraform backend blocks. It can also occur when the opposite occurs.


pkg/commands/drift/drift.go
```diff
@@ -329,9 +330,9 @@ func differenceMap(left, right map[string]*assetinventory.AssetIAM) map[string]s
 	return found
 }
 
-// differenceSet finds the keys located in the left set that are missing in the right set.
+// DifferenceSet finds the keys located in the left set that are missing in the right set.
 // We return a set so that we can do future comparisons easily with the result.
-func differenceSet(left, right map[string]struct{}) map[string]struct{} {
+func DifferenceSet(left, right map[string]struct{}) map[string]struct{} {
```
sethvargo, line 335: We've got [set functions in pkg now](https://github.com/abcxyz/pkg/blob/7ad5bd201d223baaa666fec40b1d13f4f2d571a3/sets/maps.go#L117). Do any of those work?

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -0,0 +1,205 @@
...
+		config, _, err := terraform.ParseBackendConfig(f)
+		if err != nil {
+			return fmt.Errorf("failed to parse Terraform backend config: %w", err)
+		}
+		if config.GCSBucket == nil || *config.GCSBucket == "" {
+			return fmt.Errorf("unsupported backend type for terraform config at %s - only gcs backends are supported", f)
+		}
+		gcsBuckets = append(gcsBuckets, *config.GCSBucket)
+		expectedURIs = append(expectedURIs, fmt.Sprintf("gs://%s/%s/default.tfstate", *config.GCSBucket, *config.Prefix))
+	}
```
sethvargo, line 148: Nit: it might be a better UX to process all of the entrypointBackendFiles and return a multierror (errors.Join). That would give the user all their invalid backends at once, instead of one at a time.

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -0,0 +1,205 @@
...
+
+	logger.Debug("finding statefiles in gcs bucket", "gcs_bucket")
+
+	gotURIs, err := c.terraformParser.StateFileURIs(ctx, []string{gcsBucket})
+	if err != nil {
+		return fmt.Errorf("failed to determine state file URIs for gcs bucket %s: %w", gcsBucket, err)
+	}
+
+	statefilesNotInRemote := drift.DifferenceSet(Set(expectedURIs), Set(gotURIs))
+	statefilesNotInLocal := drift.DifferenceSet(Set(gotURIs), Set(expectedURIs))
```
sethvargo, line 172: We've got slice functions that do this now: https://github.com/abcxyz/pkg/blob/7ad5bd201d223baaa666fec40b1d13f4f2d571a3/sets/slices.go#L143

pkg/terraform/terraform.go
```diff
@@ -159,6 +171,41 @@ func hasBackendConfig(path string) (bool, hcl.Diagnostics, error) {
 	return false, diags, nil
 }
 
+// ParseBackendConfig extracts the backend configuration from the backend block.
+func ParseBackendConfig(path string) (*TerraformBackendConfig, hcl.Diagnostics, error) {
```
sethvargo, line 175: Nit: `ExtractBackendConfig` might be a better name.

go.sum
```diff
@@ -145,6 +145,7 @@ github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4/go.mod h1:
 github.com/rogpeppe/fastuuid v1.2.0/go.mod h1:jVj6XXZzXRy/MSR5jhDC/2q6DgLz+nrA6LYCDYWNEvQ=
 github.com/segmentio/asm v1.2.0 h1:9BQrFxC+YOHJlTlHGkTrFWf59nbL3XnCoFLTwDCI7ys=
 github.com/segmentio/asm v1.2.0/go.mod h1:BqMnlJP91P8d+4ibuonYZw9mfnzI9HfxselHZr5aAcs=
+github.com/sergi/go-diff v1.0.0 h1:Kpca3qRNrduNnOQeazBd0ysaKrUJiIuISHxogkT9RPQ=
```
sethvargo, line 148: I'm not seeing where this is used?

sethvargo (changes requested):

go.sum
dcreey: go mod tidy pulled it in. Lint fails without it.

pkg/commands/drift/drift.go

pkg/commands/drift/drift.go
```diff
@@ -340,3 +341,13 @@ func differenceSet(left, right map[string]struct{}) map[string]struct{} {
 	}
 	return found
 }
+
+// Keys returns the sorted keys in the map.
+func Keys(m map[string]struct{}) []string {
```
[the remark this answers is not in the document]

pkg/commands/drift/statefiles/statefiles.go
dcreey: Done

pkg/commands/drift/statefiles/statefiles.go
dcreey: Done

pkg/terraform/terraform.go
dcreey: Done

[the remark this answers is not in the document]
dcreey: Done

pkg/commands/drift/drift.go
```diff
@@ -317,26 +318,11 @@ func (d *IAMDriftDetector) URI(i *assetinventory.AssetIAM) string {
...
-		if _, f := right[key]; !f {
-			found[key] = struct{}{}
-		}
+// KeySet returns the keys in the map.
+func KeySet[K comparable, V any](m map[K]V) map[K]struct{} {
+	keys := make(map[K]struct{}, len(m))
+	for k := range m {
+		keys[k] = struct{}{}
 	}
-	return found
```
sethvargo, line 341: Previously you were sorting the keys - do you want to do that here? Also https://pkg.go.dev/golang.org/x/exp/maps#Keys might be easier (also unsorted).

dcreey: I am using Keys to replace it now everywhere. Also sorted the outputs.

pkg/commands/drift/drift.go
```diff
@@ -317,26 +318,11 @@ func (d *IAMDriftDetector) URI(i *assetinventory.AssetIAM) string {
...
-// differenceSet finds the keys located in the left set that are missing in the right set.
-// We return a set so that we can do future comparisons easily with the result.
-func differenceSet(left, right map[string]struct{}) map[string]struct{} {
-	found := make(map[string]struct{})
-	for key := range left {
-		if _, f := right[key]; !f {
-			found[key] = struct{}{}
-		}
+// KeySet returns the keys in the map.
+func KeySet[K comparable, V any](m map[K]V) map[K]struct{} {
```
sethvargo, line 322: I think you can delete this now?

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -0,0 +1,214 @@
...
+	return nil
+}
+
+func Set(values []string) map[string]struct{} {
+	set := make(map[string]struct{})
+	for _, v := range values {
+		set[v] = struct{}{}
+	}
+	return set
+}
```
sethvargo, line 196: I'm only seeing this used above on L178-179 where we're converting slices to maps. Would it be better to subtract the slices directly and drop this function?

```go
	statefilesNotInRemote := sets.Subtract(expectedURIs, gotURIs)
	statefilesNotInLocal := sets.Subtract(gotURIs, expectedURIs)
```

Then you could also avoid `maps.Keys` and `sort.Strings` on L201-202 and L209-210.

dcreey: I still want the sort so that the output to the user is in a reliable order. Updated the rest

pkg/commands/drift/drift.go
dcreey: nvm. Removed it.

sethvargo (approved):

PR #168, opened 2023-07-26
feat(statefiles): Use github issue for tracking statefile drift

dcreey:


pkg/commands/drift/github.go
```diff
@@ -17,92 +17,70 @@ package drift
...
-
-        See the comment(s) below to see details of the drift
-
-        Please determine which parts are correct, and submit updated
-        terraform config and/or remove the extra policies.
-
-        Re-run drift detection manually once complete to verify all diffs are properly resolved.`
-)
+type GitHubDriftIssueService struct {
+	GH         *github.GitHubClient
```
sethvargo, line 26: Style nit: it'd be better to name this GitHubClient or just Client. Also, does it need to be public?

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -35,18 +37,32 @@ import (
...
 
 	flags.GitHubFlags
 	flags.RetryFlags
+	driftflags.DriftIssueFlags
 
 	flagGCSBucket string
 
-	githubClient    github.GitHub
 	terraformParser parser.Terraform
+	issueService    drift.GitHubDriftIssueService
```
sethvargo, line 65: This should be a pointer since it's a struct

pkg/commands/drift/github.go
dcreey: Will update the name. Regarding it being public - I want to use the Service outside of the drift pkg (in statefiles). Precedent would lead me to believe that we should create a constructor. In this case I thought adding the constructor only adds an unnecessary function + call to function (one with a lot of string arguments). wdyt?

verbanicm: I believe he is referring to `GH` -> `githubClient` so its private to the struct, not the GitHubDriftIssueService. I think its fine that GitHubDriftIssueService is public and reusable. If its common enough can we move it into the `github` package instead (or make it generic enough to move it in there)?

dcreey: Right right. But I don't have a constructor for the  GitHubDriftIssueService. So in order to create an instance of it you need to set the necessary private member variables (doesn't work from another pkg if they are private).

dcreey: Done

pkg/commands/drift/statefiles/statefiles.go
dcreey: Done

pkg/commands/drift/flags/driftissue.go
```diff
@@ -0,0 +1,62 @@
...
+// DriftIssueFlags represent the shared drift github issue flags among all commands.
+// Embed this struct into any commands that needs to create drift GitHub issues.
+type DriftIssueFlags struct {
+	FlagSkipGitHubIssue            bool
+	FlagGitHubIssueLabels          []string
+	FlagGitHubIssueAssignees       []string
+	FlagGitHubCommentMessageAppend string
+}
+
+func (d *DriftIssueFlags) AddFlags(set *cli.FlagSet, defaultIssueLabel string) {
```
sethvargo, line 30: This API is a little bit weird. I wonder if we should accept a struct instead?

sethvargo (approved):

pkg/commands/drift/flags/driftissue.go
dcreey: Done

sethvargo (approved):

PR #188, opened 2023-08-18
chore: refactor the part of init into entrypoints

verbanicm:
The command `initialize` seems to being doing too much and is confusing. It currently mixes core logic (finding entrypoints) with GitHub specific logic (deleting PR comments, required permissions, etc).

This PR is one of multiple PRs that will split this logic apart.

In this PR, the logic related to determining the Guardian entrypoints is moved to its own command. The follow up PRs will create a new command for handling GitHub specific logic.

### DIFF
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -1,0 +1,251 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package entrypoints provides the functionality to determine the entrypoints for Guardian.
+package entrypoints
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/abcxyz/guardian/pkg/flags"
+	"github.com/abcxyz/guardian/pkg/git"
+	"github.com/abcxyz/guardian/pkg/terraform"
+	"github.com/abcxyz/guardian/pkg/util"
+	"github.com/abcxyz/pkg/cli"
+	"github.com/abcxyz/pkg/logging"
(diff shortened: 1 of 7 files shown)


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -0,0 +1,251 @@
...
+
+	logger.DebugContext(ctx, "Starting Guardian init")
+
+	if c.flagFormat == "" {
+		c.flagFormat = "text"
+	}
+
+	if _, ok := allowedFormats[c.flagFormat]; !ok {
+		return fmt.Errorf("invalid format flag: %s (supported formats are: %s)", c.flagFormat, strings.Join(maps.Keys(allowedFormats), ", "))
+	}
```
sethvargo, line 162: It would be better to do this kind of validation/defaulting in [AfterParse](https://github.com/abcxyz/pkg/blob/9b36be5112d6211f33d81a8ebcb29693e66a3809/cli/command_doc_after_parse_test.go#L66-L78). Then you can easily test it too.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -0,0 +1,251 @@
...
+	f.BoolVar(&cli.BoolVar{
+		Name:   "skip-detect-changes",
+		Target: &c.flagSkipDetectChanges,
+		Usage:  "Skip detecting file changes and run for all entrypoint directories.",
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:   "format",
+		Target: &c.flagFormat,
+		Usage:  fmt.Sprintf("The format to print the output directories. The supported formats are: %s.", strings.Join(maps.Keys(allowedFormats), ", ")),
```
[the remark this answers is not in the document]
sethvargo: But actually, I suggest making `var allowFormatNames = maps.Keys(allowedFormats)` a global variable. Because you also probably want to add a predictor here with those values as a list.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -0,0 +1,251 @@
...
+	case "json":
+		if err := json.NewEncoder(c.Stdout()).Encode(dirs); err != nil {
+			return fmt.Errorf("failed to create json string: %w", err)
+		}
+	case "text":
+		for _, dir := range dirs {
+			c.Outf("%s", dir)
+		}
+	default:
+		return fmt.Errorf("invalid format flag: %s (supported formats are: %s)", c.flagFormat, strings.Join(maps.Keys(allowedFormats), ", "))
```
sethvargo, line 247: I think the string list should be moved to a private global var, given the frequency of use.

pkg/commands/entrypoints/entrypoints.go
verbanicm: Ah, nice!

yolocs (approved):

PR #192, opened 2023-08-23
feat: migrate plan-status to go

verbanicm:
Migrate the existing plan-status action to the workflows command. This command is used by a github workflow as a post status job. The users can configured required status checks to ensure this check is required because when using a matrix strategy, the plan job has dynamic names and makes it impossible to have required status checks for all possible directories.

This change also cleans up required flags for several commands and adds tests for them.

### DIFF
--- a/pkg/commands/workflows/plan_status_comments.go
+++ b/pkg/commands/workflows/plan_status_comments.go
@@ -1,0 +1,225 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


pkg/commands/workflows/plan_status_comments.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// this case improves user experience, when the planning job does not have a plan diff
+	// there are no comments, this informs the user that the job ran successfully
+	if c.flagInitResult == "success" && c.flagPlanResult == "success" {
+		if _, err := c.gitHubClient.CreateIssueComment(
+			ctx,
+			c.GitHubFlags.FlagGitHubOwner,
+			c.GitHubFlags.FlagGitHubRepo,
+			c.flagPullRequestNumber,
+			fmt.Sprintf("%s 🟩 Plan completed successfully. %s", plan.CommentPrefix, c.gitHubLogURL),
```
yolocs, line 178: I thought we print out the plan diff?

yolocs (approved):

pkg/commands/workflows/plan_status_comments.go
verbanicm: We do that in the plan step, this is a follow up job. See the PR description.

PR #193, opened 2023-08-24
chore: import template workflows from gcp-org-terraform-template

gjonathanhong:
…mplates


abc.templates/README.md
```diff
@@ -0,0 +1,17 @@
...
+```shell
+abc templates render \
+  -input=guardian_wif_provider=<WIF_PROVIDER> \
+  -input=guardian_service_account=<SERVICE_ACCOUNT> \
+  -input=guardian_state_bucket=<GUARDIAN_STATE_BUCKET> \
+  github.com/abcxyz/guardian.git//abc.templates/workflows
+```
+
+#### Optional inputs:
+- `terraform_version`: Defaults to `1.5.6`.
```
verbanicm, line 16: I believe we can only currently support 1.5.4 because of token minter and its version checksums: https://github.com/abcxyz/secure-setup-terraform/blob/v0.2.7/terraform-checksums.json

abc.templates/workflows/contents/guardian-admin.yml
verbanicm, line 1: I wonder if we should name this something other than "workflows", what if there are multiple workflow tempaltes, maybe this should be base-workflows, or default-workflows

verbanicm:
Rebase

verbanicm (approved):

PR #202, opened 2023-09-07
feat: allow skipping recursion on entrypoints

dcreey:


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -130,7 +131,18 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 		Default: false,
 	})
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "no-recursion",
```
verbanicm, line 135: Not sure im a fan of the naming, `skip-nested-entrypoints`? Something with skip to keep naming similar.

dcreey: Done

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -130,7 +131,18 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
...
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "skip-nested-entrypoints",
+		Target:  &c.flagSkipNestedEntrypoints,
+		Usage:   `Whether or not recurse and find entrypoints nested beneath the target entrypoint.`,
+		Default: false,
+	})
+
 	set.AfterParse(func(existingErr error) (merr error) {
+		if c.flagSkipNestedEntrypoints && !c.flagSkipDetectChanges {
```
verbanicm, line 142: I am not sure I agree with this API, maybe instead we should check on L210 if this is true and just skip detect changes.

dcreey: I'm not a big fan of APIs that change default options for other flags depending on other flags you've set. It should be explicit. The user should know have to explicitly set each flag that impacts overall behavior.

If we find ourselves with a lot of flags that would lead to a complex matrix of options then we should consider whether or not we should split the api into several subcommands.

drevell (approved):
LGTM for readability

pkg/commands/entrypoints/entrypoints.go
verbanicm: Fair.

verbanicm (dismissed):

pkg/commands/entrypoints/entrypoints.go
verbanicm: Maybe we should re-think this and instead of "skipping" we should enable. IE, detect-changes and nested-entrypoints, to force the "on" behavior and it defaults to off.

dcreey: imo the defaults should be the mostly likely path for the user. I think most users will want detect-changes on and nested-entrypoints on. What are the use cases?

* As a user I would like to find all terraform entrypoints in a given directory and its subdirectories
* As a user I would like to find all terraform entrypoints that have changed in a directory and its subdirectories
* As a user I would like to check if a given directory has an entrypoint
* As a user I would like to check if a given directory has an entrypoint that has changed

Now that I spell it out these kind of seem like different APIs. Maybe we should have a validate-entrypoint and entrypoints subcommand?

dcreey: Updated based on offline conversation.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -107,8 +108,8 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 	})
 
 	f.BoolVar(&cli.BoolVar{
-		Name:   "skip-detect-changes",
-		Target: &c.flagSkipDetectChanges,
+		Name:   "detect-changes",
+		Target: &c.flagDetectChanges,
```
verbanicm, line 112: Update usage string

dcreey: Done

drevell (approved):

PR #203, opened 2023-09-07
feat: allow running statefile drift on repo with different backends

dcreey:
Some infra repos store state in multiple different buckets. Statefile drift detection should find all uris in all buckets referenced in terraform configuration.

Tested it against X's repo for their gcp-org which has many backend buckets.

### DIFF
--- a/pkg/commands/drift/statefiles/statefiles.go
+++ b/pkg/commands/drift/statefiles/statefiles.go
@@ -23,6 +23,11 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/abcxyz/pkg/cli"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/sets"
+	"golang.org/x/exp/maps"
+
 	"github.com/abcxyz/guardian/pkg/commands/drift"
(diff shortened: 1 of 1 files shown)


pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -163,33 +154,19 @@ func (c *DriftStatefilesCommand) Process(ctx context.Context) error {
...
-		}
-	}
-
-	if gcsBucket == "" {
-		return fmt.Errorf("unable to determine gcs bucket - please provide the gcsBucket flag or point at a terraform config with gcs backends")
-	}
-
-	logger.DebugContext(ctx, "finding statefiles in gcs bucket",
-		"gcs_bucket", gcsBucket)
+	logger.DebugContext(ctx, "finding statefiles in gcs buckets",
```
yolocs, line 164: nit: `gcsBuckets` is a map and the output will look like "map[bar:{} foo:{}]". Instead, right before this line, we could turn it into a slice and just use the slice below.

dcreey: Done

yolocs (approved):

PR #204, opened 2023-09-08
feat: update statefile drift to check all repos and buckets

dcreey:
These change updates statefile drift so that it 

* Clones all github repositories that have a given set of topics (e.g. terraform, guardian) - this will be limited by the github token passed - so if you only have access to 1 repo you will only get 1 repo.
* (noop) Determines all expected statefile locations in gcs using the method already used
* Fetches all gcs buckets - this will be limited by the access that you/your SA has.
* (noop) Determines all actual statefile locations in gcs using the method already used

To support environments that don't have elevated github permissions I introduced a new flag called `detect-gcs-buckets-from-terraform` that will not rely on fetching GCS buckets from the asset inventory API. Instead it will parse them from the terraform configuration (this is how it used to behave before this change).

### DIFF
--- a/pkg/commands/drift/statefiles/statefiles.go
+++ b/pkg/commands/drift/statefiles/statefiles.go
@@ -28,13 +28,16 @@ import (
 	"github.com/abcxyz/pkg/sets"
 	"golang.org/x/exp/maps"
 
+	"github.com/abcxyz/guardian/pkg/assetinventory"
 	"github.com/abcxyz/guardian/pkg/commands/drift"
 	driftflags "github.com/abcxyz/guardian/pkg/commands/drift/flags"
 	"github.com/abcxyz/guardian/pkg/flags"
+	"github.com/abcxyz/guardian/pkg/git"
 	"github.com/abcxyz/guardian/pkg/github"
 	"github.com/abcxyz/guardian/pkg/terraform"
 	"github.com/abcxyz/guardian/pkg/terraform/parser"
 	"github.com/abcxyz/guardian/pkg/util"
+	githubAPI "github.com/google/go-github/v53/github"
 )
 
 var _ cli.Command = (*DriftStatefilesCommand)(nil)
@@ -61,8 +64,17 @@ type DriftStatefilesCommand struct {
 	flags.RetryFlags
 	driftflags.DriftIssueFlags
 
-	terraformParser parser.Terraform
-	issueService    *drift.GitHubDriftIssueService
+	flagOrganizationID                string
+	flagGitHubRepoQuery               string
+	flagGCSBucketQuery                string
+	flagDetectGCSBucketsFromTerraform bool
+	flagTerraformRepoTopics           []string
+
+	assetInventoryClient assetinventory.AssetInventory
+	gitClient            git.Git
+	githubClient         github.GitHub
+	issueService         *drift.GitHubDriftIssueService
+	terraformParser      parser.Terraform
 }
 
 func (c *DriftStatefilesCommand) Desc() string {
@@ -86,6 +98,43 @@ func (c *DriftStatefilesCommand) Flags() *cli.FlagSet {
 		DefaultIssueLabel: "guardian-statefile-drift",
 	})
 
+	// Command options
+	f := set.NewSection("COMMAND OPTIONS")
+
+	f.StringVar(&cli.StringVar{
+		Name:    "organization-id",
+		Target:  &c.flagOrganizationID,
+		Example: "123435456456",
+		Usage:   `The Google Cloud organization ID for which to detect drift.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "github-repo-query",
+		Target:  &c.flagGitHubRepoQuery,
+		Example: "labels.terraform:*",
+		Usage:   `The label to use to find GCS buckets with Terraform statefiles.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "gcs-bucket-query",
+		Target:  &c.flagGCSBucketQuery,
+		Example: "labels.terraform:*",
+		Usage:   `The label to use to find GCS buckets with Terraform statefiles.`,
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:   "detect-gcs-buckets-from-terraform",
+		Target: &c.flagDetectGCSBucketsFromTerraform,
+		Usage:  `Whether or not to use the terraform backend configs to determine gcs buckets.`,
+	})
+
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "github-repo-terraform-topics",
+		Target:  &c.flagTerraformRepoTopics,
+		Example: "terraform,guardian",
+		Usage:   `Topics to use to identify github repositories that contain terraform configurations.`,
+	})
+
 	return set
 }
 
@@ -107,8 +156,10 @@ func (c *DriftStatefilesCommand) Run(ctx context.Context, args []string) error {
 	}
 	c.directory = dirAbs
 
+	c.gitClient = git.NewGitClient(c.directory)
+	c.githubClient = github.NewClient(ctx, c.GitHubFlags.FlagGitHubToken)
 	c.issueService = drift.NewGitHubDriftIssueService(
-		github.NewClient(ctx, c.GitHubFlags.FlagGitHubToken),
+		c.githubClient,
 		c.GitHubFlags.FlagGitHubOwner,
 		c.GitHubFlags.FlagGitHubRepo,
 		issueTitle,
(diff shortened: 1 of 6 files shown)


pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -86,6 +98,43 @@ func (c *DriftStatefilesCommand) Flags() *cli.FlagSet {
...
+		Example: "123435456456",
+		Usage:   `The Google Cloud organization ID for which to detect drift.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "github-repo-query",
+		Target:  &c.flagGitHubRepoQuery,
+		Example: "labels.terraform:*",
+		Usage:   `The label to use to find GCS buckets with Terraform statefiles.`,
+	})
```
verbanicm, line 116: Is this used? Seems like logic switch to topics instead of labels.

dcreey: Removed

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -131,6 +178,25 @@ func (c *DriftStatefilesCommand) Process(ctx context.Context) error {
...
+	}
+	repositoriesWithTerraform := []*github.Repository{}
+	for _, r := range repositories {
+		if len(sets.Subtract(r.Topics, c.flagTerraformRepoTopics)) == 0 {
+			repositoriesWithTerraform = append(repositoriesWithTerraform, r)
+		}
+	}
+
+	for _, r := range repositoriesWithTerraform {
+		if err = c.gitClient.CloneRepository(ctx, c.GitHubFlags.FlagGitHubToken, r.Owner, r.Name); err != nil {
```
verbanicm, line 194: Seems like this clones the code into the CWD, there COULD be potential collisions with sub-folder names. We should try to target a different directory somehow.

pkg/git/git.go
```diff
@@ -77,6 +79,28 @@ func (g *GitClient) DiffDirsAbs(ctx context.Context, sourceRef, destRef string)
...
+	logger := logging.FromContext(ctx).With("working_dir", g.workingDir)
+
+	var stdout, stderr bytes.Buffer
+
+	_, err := child.Run(ctx, &child.RunConfig{
+		Stdout:     &stdout,
+		Stderr:     &stderr,
+		WorkingDir: g.workingDir,
+		Command:    "git",
+		Args:       []string{"clone", fmt.Sprintf("https://%s@github.com/%s/%s.git", githubToken, owner, repo)},
```
verbanicm, line 93: Should we use `--single-branch` and `--no-tags` just for efficiency. We dont need everything just everything from the default branch right?

pkg/github/github.go
```diff
@@ -79,6 +88,9 @@ const (
 
 // GitHub provides the minimum interface for sending requests to the GitHub API.
 type GitHub interface {
+	//  ListRepositories all repositories and returns details about the repositories.
```
verbanicm, line 91:
```suggestion
	//  ListRepositories lists all repositories and returns details about the repositories.
```

pkg/github/github.go
```diff
@@ -146,6 +158,53 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+// ListRepositories all repositories and returns details about the repositories.
+func (g *GitHubClient) ListRepositories(ctx context.Context, owner string, opts *github.RepositoryListByOrgOptions) ([]*Repository, error) {
+	pageStart := func(i *int) bool { return i == nil }
+	pageEnd := func(i *int) bool { return *i == 0 }
+	uniqueResponses := make(map[int64]*Repository)
+	opt := *opts
+
+	var page *int // Use nil to indicate start of request.
+
+	for pageStart(page) || !pageEnd(page) {
```
verbanicm, line 170: Why are we not paginating like the example: https://github.com/google/go-github?tab=readme-ov-file#pagination

This seems overly complex.

pkg/github/github.go
```diff
@@ -146,6 +158,53 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
 	return g
 }
 
+// ListRepositories all repositories and returns details about the repositories.
```
verbanicm, line 161:
```suggestion
// ListRepositories lists all repositories and returns details about the repositories.
```

pkg/github/github.go
```diff
@@ -146,6 +158,53 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+			repos, resp, err := g.client.Repositories.ListByOrg(ctx, owner, &opt)
+			if err != nil {
+				if _, ok := ignoredStatusCodes[resp.StatusCode]; !ok {
+					return retry.RetryableError(err)
+				}
+
+				return fmt.Errorf("failed to list issues: %w", err)
+			}
+
+			for _, r := range repos {
```
verbanicm, line 184: Whats the use case where we get duplicate repos?

pkg/github/github.go
dcreey: Because we want to use the retry function to retry certain errors. It adds a little bit of complexity here.

pkg/github/github.go
dcreey: https://github.com/abcxyz/guardian/pull/134#discussion_r1265598622

I will add a comment to justify this.

pkg/github/github.go
verbanicm: ack per offline comment

pkg/github/github.go
verbanicm: got it, thanks!

pkg/commands/drift/statefiles/statefiles.go
dcreey: My goal is to document this and let the user decide. Then they can `DIR=$(mktmp) && guardian ... $DIR`.

We could instead clone to a tmp dir in the tool. wdyt?

pkg/git/git.go
dcreey: Done

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -86,6 +101,55 @@ func (c *DriftStatefilesCommand) Flags() *cli.FlagSet {
...
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:   "detect-gcs-buckets-from-terraform",
+		Target: &c.flagDetectGCSBucketsFromTerraform,
+		Usage:  `Whether or not to use the terraform backend configs to determine gcs buckets.`,
+	})
+
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "github-repo-terraform-topics",
```
yolocs, line 128: Should we make sure this flag is not empty? Or having a default

pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -129,20 +199,54 @@ func (c *DriftStatefilesCommand) Process(ctx context.Context) error {
 		With("github_repo", c.GitHubFlags.FlagGitHubOwner)
 
 	logger.DebugContext(ctx, "starting Guardian drift statefiles")
+
+	// Clone all git repositories.
+	repositories, err := c.githubClient.ListRepositories(ctx, c.GitHubFlags.FlagGitHubOwner, &githubAPI.RepositoryListByOrgOptions{})
```
yolocs, line 204: This func is now 120 lines and getting a little hard to read even with the clear inline documentation you have. Is it possible to encapsulate some sub-logic into separate functions?

pkg/github/github.go
```diff
@@ -146,6 +158,57 @@ func NewClient(ctx context.Context, token string, opts ...Option) *GitHubClient
...
+			if page != nil {
+				opt.Page = *page
+			}
+			repos, resp, err := g.client.Repositories.ListByOrg(ctx, owner, &opt)
+			if err != nil {
+				if _, ok := ignoredStatusCodes[resp.StatusCode]; !ok {
+					return retry.RetryableError(err)
+				}
+
+				return fmt.Errorf("failed to list issues: %w", err)
```
yolocs, line 185:
```suggestion
				return fmt.Errorf("failed to list repositories: %w", err)
```

yolocs (approved):

pkg/commands/drift/statefiles/statefiles.go
verbanicm: I think the tmp dir is better, does that affect your ignore dir logic?

verbanicm: This can be a follow up.

verbanicm (approved):

pkg/commands/drift/statefiles/statefiles.go
dcreey: It does affect it. I'll do it in a follow up.

dcreey: https://github.com/abcxyz/guardian/issues/206

PR #205, opened 2023-09-11
chore: indicate unknown parent in logs and issue

dcreey:
When a parent project or folder cannot be identified we should indicate that

### DIFF
--- a/pkg/commands/drift/drift.go
+++ b/pkg/commands/drift/drift.go
@@ -325,7 +325,9 @@ func (d *IAMDriftDetector) URI(i *assetinventory.AssetIAM) string {
 			resourceName = p.Name
 		}
 		return fmt.Sprintf("/organizations/%s/projects/%s/%s/%s", d.organizationID, resourceName, role, i.Member)
-	} else {
+	} else if i.ResourceType == assetinventory.Organization {
 		return fmt.Sprintf("/organizations/%s/%s/%s", d.organizationID, role, i.Member)
+	} else {
+		return fmt.Sprintf("unknownParent:/organizations/%s/%s/%s/%s/%s", d.organizationID, i.ResourceType, i.ResourceID, role, i.Member)
 	}
 }
--- a/pkg/terraform/parser/parser.go
+++ b/pkg/terraform/parser/parser.go
@@ -22,6 +22,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/abcxyz/pkg/logging"
+
 	"github.com/abcxyz/guardian/pkg/assetinventory"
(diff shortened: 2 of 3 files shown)


verbanicm (approved):

verbanicm (dismissed):

pkg/commands/drift/drift.go
```diff
@@ -325,7 +325,9 @@ func (d *IAMDriftDetector) URI(i *assetinventory.AssetIAM) string {
 			resourceName = p.Name
 		}
 		return fmt.Sprintf("/organizations/%s/projects/%s/%s/%s", d.organizationID, resourceName, role, i.Member)
-	} else {
+	} else if i.ResourceType == assetinventory.Organization {
```
yolocs, line 328: nit: It would be slightly clearer if we use `switch`

```go
switch i.ResourceType {
case assetinventory.Folder:
  //...
case assetinventory.Project:
  //...
case assetinventory.Organization:
  //...
default:
  return fmt.Sprintf("unknownParent:/organizations/%s/%s/%s/%s/%s", d.organizationID, i.ResourceType, i.ResourceID, role, i.Member)
}
```

pkg/terraform/parser/parser.go
```diff
@@ -177,12 +176,16 @@ func (p *TerraformParser) parseIAMBindingForOrg(instances []ResourceInstance) []
...
-func (p *TerraformParser) parseIAMBindingForFolder(instances []ResourceInstance) []*assetinventory.AssetIAM {
+func (p *TerraformParser) parseIAMBindingForFolder(ctx context.Context, instances []ResourceInstance) []*assetinventory.AssetIAM {
 	var iams []*assetinventory.AssetIAM
 	for _, i := range instances {
 		for _, m := range i.Attributes.Members {
 			folderID := strings.TrimPrefix(i.Attributes.Folder, "folders/")
 			parentID, parentType := p.maybeFindGCPAssetIDAndType(folderID)
+			if parentType == assetinventory.Unknown {
+				logger := logging.FromContext(ctx)
+				logger.Warn("failed to locate GCP folder - is this folder deleted?", "folder", folderID)
```
yolocs, line 187:
```suggestion
				logger.WarnContext(ctx, "failed to locate GCP folder - is this folder deleted?", "folder", folderID)
```

yolocs: Throughout

pkg/commands/drift/drift.go

pkg/terraform/parser/parser.go
dcreey: Done

PR #209, opened 2023-09-14
fix: duplicate flag values

dcreey:


verbanicm (approved):

pkg/commands/drift/flags/driftissue.go
```diff
@@ -54,7 +50,7 @@ func (d *DriftIssueFlags) Register(set *cli.FlagSet, options *Options) {
 		Target:  &d.FlagGitHubIssueLabels,
 		Example: "guardian-iam-drift",
 		Usage:   `The labels to use on any created GitHub Issues.`,
-		Default: []string{options.DefaultIssueLabel},
+		Default: []string{},
```
verbanicm, line 53: Should this be removed and do a nil check instead? What if someone wants zero labels to be given?

dcreey: That would be problematic as we manage the issues based on these labels (lookup, close, etc). I wouldn't want to add that complexity to this tool

drevell (commented):
Holding off on review until Mike's comment is addressed

dcreey:
Its been addressed

sethvargo (approved):

PR #226, opened 2023-10-31
Use the head.sha instead of head.ref

mering:
Using `head.ref`/`base.ref` can lead to race conditions where the branch (ref) is modified after a workflow is started which results in either the wrong revision to be checked out or fails when the branch is deleted.

### DIFF
--- a/abc.templates/default-workflows/contents/guardian-plan.yml
+++ b/abc.templates/default-workflows/contents/guardian-plan.yml
@@ -51,7 +51,7 @@ jobs:
         uses: 'actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9' # ratchet:actions/checkout@v3
         with:
           fetch-depth: 0 # get everything so we can use git diff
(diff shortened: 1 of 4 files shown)


abc.templates/default-workflows/contents/guardian-plan.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: 'actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9' # ratchet:actions/checkout@v3
         with:
           fetch-depth: 0 # get everything so we can use git diff
-          ref: '${{ github.event.pull_request.head.ref }}'
+          ref: '${{ github.event.pull_request.head.sha }}'
```
dcreey, line 54: I think you should update the usages in the .github/workflow/ files too

mering: Done.

verbanicm:
Closing this as permissions from fork dont work with pull request event:  https://github.com/abcxyz/guardian/pull/227

PR #228, opened 2023-11-16
fix: add formatting check to guardian plan

verbanicm:
First push should have formatting failure, will fix for merge.

### DIFF
--- a/pkg/terraform/format.go
+++ b/pkg/terraform/format.go
@@ -1,0 +1,73 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package terraform
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/abcxyz/guardian/pkg/util"
+)
+
+// FormatOptions are the set of options for running terraform format.
+type FormatOptions struct {
+	Check     *bool
+	Diff      *bool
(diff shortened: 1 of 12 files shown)


pkg/terraform/format.go
```diff
@@ -0,0 +1,71 @@
...
+
+// formatArgsFromOptions generated the terrafrom format arguments from the provided options.
+func formatArgsFromOptions(opts *FormatOptions) []string {
+	args := make([]string, 0, 6) // 6 potential args to be added
+
+	if opts == nil {
+		return args
+	}
+
+	if opts.Check != nil && *opts.Check {
```
sethvargo, line 41: It might be helpful to make this a generic function like `boolVal`

```go
func boolVal(b *bool) bool {
  if v == nil {
    return false
  }
  return *v
}
```

of if you wanted to be really fancy:

```go
func boolVal(b *bool) (val, ok bool) {
  if v == nil {
    return
  }
  return *v, true
}
```

pkg/terraform/terraform_mock.go
```diff
@@ -83,6 +84,15 @@ func (m *MockTerraformClient) Show(ctx context.Context, stdout, stderr io.Writer
 	return 0, nil
 }
 
+func (m *MockTerraformClient) Format(ctx context.Context, stdout, stderr io.Writer, opts *FormatOptions) (int, error) {
+	if m.FormatResponse != nil {
```
sethvargo, line 88: why would this be nil?

verbanicm: If someone didnt configure it in their test, e.g. they didnt care about the response so just return some default instead.

pkg/terraform/format.go
verbanicm: I prefer option 1, will update

pkg/util/util.go
```diff
@@ -27,6 +27,14 @@ func Ptr[T any](v T) *T {
 	return &v
 }
 
+// BoolVal returns the value of a bool pointer or false if it is nil.
+func BoolVal(v *bool) bool {
```
sethvargo, line 31: You could make this generic I guess:

```go
// PtrVal returns the value of the given pointer. If the provided input is nil,
// it returns the zero value for the pointer's underlying type.
func PtrVal[T any](v *T) T {
  if v == nil {
    v = new(T)
  }
  return *v
}
```

PR #234, opened 2023-12-07
feat: designate which statefiles are empty in drift detection

dcreey:
This enables statefiles drift detection to inform the user which statefiles that were found are actually empty (result of an init without an apply)

Example of new output:
```
Found state locally that are not in remote
> gs://terraform-abc123/terraform/a/default.tfstate
> gs://terraform-abc123/terraform/b/default.tfstate

Found statefiles in remote that are not in local           
> gs://terraform-abc123/terraform/c/default.tfstate
> gs://terraform-abc123/terraform/d/default.tfstate
                                                                                                                          
Found empty statefiles in remote that are not in local
> gs://terraform-abc123/terraform/e/default.tfstate
> gs://terraform-abc123/terraform/f/default.tfstate
```

### DIFF
--- a/pkg/commands/drift/statefiles/statefiles.go
+++ b/pkg/commands/drift/statefiles/statefiles.go
@@ -381,7 +381,7 @@ func (c *DriftStatefilesCommand) emptyStateFiles(ctx context.Context, gcsURIs []
 			if err != nil {
 				return nil, fmt.Errorf("failed to get determine if state file URI has resources: %w", err)
 			}
-			return &Resource{*empty, uri}, nil
+			return &Resource{empty, uri}, nil
 		}); err != nil && !errors.Is(err, workerpool.ErrStopped) {
 			return nil, fmt.Errorf("failed to execute terraform resources task: %w", err)
 		}
@@ -393,11 +393,11 @@ func (c *DriftStatefilesCommand) emptyStateFiles(ctx context.Context, gcsURIs []
 	}
 
 	var emptyStateFiles []string
-	errs := []error{}
+	var merr error
 	for _, r := range results {
 		if err := r.Error; err != nil {
 			if !errors.Is(err, workerpool.ErrStopped) {
-				errs = append(errs, fmt.Errorf("failed to execute resource task: %w", err))
+				merr = errors.Join(merr, fmt.Errorf("failed to execute resource task: %w", err))
 			}
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/commands/drift/statefiles/statefiles.go
```diff
@@ -353,6 +365,52 @@ func (c *DriftStatefilesCommand) actualStatefileUris(ctx context.Context, logger
...
+		}
+	}
+
+	results, err := w.Done(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute terraform resources tasks in parallel: %w", err)
+	}
+
+	var emptyStateFiles []string
+	errs := []error{}
```
yolocs, line 396: Any special reason to prefer a slice instead of just joining errors directly. Like

```go
var merr error
for _, r := range results {
  // ...
  merr = errors.Join(merr, err)
}
if merr != nil {
  return nil, fmt.Errorf("failed to execute terraform resource tasks in parallel: %w", merr)
}
```

pkg/terraform/parser/parser.go
```diff
@@ -115,6 +122,26 @@ func (p *TerraformParser) StateFileURIs(ctx context.Context, gcsBuckets []string
 	return gcsURIs, nil
 }
 
+// ProcessStates finds all IAM in memberships, bindings, or policies in the given terraform state files.
+func (p *TerraformParser) StateWithoutResources(ctx context.Context, uri string) (*bool, error) {
```
yolocs, line 126:
```suggestion
// StateWithoutResources finds state without resources.
func (p *TerraformParser) StateWithoutResources(ctx context.Context, uri string) (*bool, error) {
```

yolocs, line 126: Weird, why `*bool`?

pkg/terraform/parser/parser.go
```diff
@@ -115,6 +121,26 @@ func (p *TerraformParser) StateFileURIs(ctx context.Context, gcsBuckets []string
 	return gcsURIs, nil
 }
 
+// StateWithoutResources determines if the given statefile at the uri contains any resources or not.
+func (p *TerraformParser) StateWithoutResources(ctx context.Context, uri string) (bool, error) {
```
verbanicm, line 125: Why isnt this functionality part of the process states logic? Seems like a wasted download call here?

yolocs (approved):

pkg/terraform/parser/parser.go
dcreey: Because ProcessStates is actually only concerned with IAM state. That could change if we make drift detection do more than just IAM. We don't use ProcessStates for statefile drift detection. Currently, we only use GCS to check which files actually exist. This adds functionality to check if the statefile has NO state.

Perhaps we should rename ProcessStates to ProcessIAMStates or something? I can do it in a follow up

verbanicm (approved):

PR #235, opened 2023-12-07
Inject custom user-agent headers

sethvargo:
This also sets us up for a future world where we can allow/deny specific environment variables from the parent process (such as forbidding users to override the path to a terraformrc file).

### DIFF
--- a/pkg/child/child.go
+++ b/pkg/child/child.go
@@ -21,10 +21,21 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 	"time"
 
+	"github.com/abcxyz/guardian/internal/version"
 	"github.com/abcxyz/pkg/logging"
 )
+
+// overrideEnvVars are the environment variables to inject into the child
+// process, no matter what the user configured. These take precedence over all
+// other configurables.
+var overrideEnvVars = []string{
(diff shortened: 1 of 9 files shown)


pkg/child/child.go
```diff
@@ -90,3 +112,41 @@ func Run(ctx context.Context, cfg *RunConfig) (int, error) {
...
+	for _, v := range osEnv {
+		k := strings.SplitN(v, "=", 2)[0]
+		if (len(allowedKeys) > 0 && anyGlobMatch(k, allowedKeys)) &&
+			!anyGlobMatch(k, deniedKeys) {
+			finalEnv = append(finalEnv, v)
+		}
+	}
+
+	// Add overrides at the end, after any filtering
+	finalEnv = append(finalEnv, overrideEnv...)
```
yolocs, line 132: What happens if overrideEnv keys appear in the user provided env vars?

yolocs (approved):

pkg/child/child.go
sethvargo: The later keys in the slice take precedence (as expected). That's taken care of by `cmd`. 

That's actually true in unix/linux in general. It's possible to have multiple keys with different values in `env` - it's why env is a slice and not a map.

PR #236, opened 2023-12-18
fix: allow any env vars for now

verbanicm:
PR #235 broke existing functionality by not allowing any env vars. Existing logic looks for application default credentials in the environment to access Terraform state backend.

For now, we will allow all environment variables and deny only one.

cc: @sethvargo, @dcreey


sethvargo:
I think https://github.com/abcxyz/guardian/pull/237 is the correct solution.

dcreey (approved):

verbanicm:
Closing in favor of #237

pkg/child/child_test.go
```diff
@@ -201,6 +201,14 @@ func TestEnviron(t *testing.T) {
 			overrideEnv: []string{},
 			exp:         []string{},
 		},
+		{
```
dcreey, line 204: I think that we should could guardian changes with workflow executions to at least plan a real terraform change.

PR #248, opened 2024-01-04
chore: document Guardian cli

dcreey:
Preview: https://github.com/abcxyz/guardian/blob/dcreey/iam-docs/cli.md

Will follow up with another PR to document the .driftignore file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@ Guardian is a Terraform actuation and enforcement tool using GitHub actions.
 
 **This is not an official Google product.**
 
+## Guardian CLI
+
+This is the underlying tool written in Golang that is enables all features. For more
+details and to understand how to use the cli see [Gaurdian CLI](./cli.md).
+
 ## Guardian Features
 
 > [!NOTE]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,6 +6,11 @@ Guardian is a Terraform actuation and enforcement tool using GitHub actions.
 
 **This is not an official Google product.**
 
+## Guardian CLI
+
+This is the underlying tool written in Golang that is enables all features. For more
+details and to understand how to use the cli see [Gaurdian CLI](./cli.md).
```
mattrandallbecker, line 12:
```suggestion
details and to understand how to use the cli see [Guardian CLI](./cli.md).
```

sethvargo:
You can also click this button to render:

screenshot-20240104-qRnW7vLz@2x

This is fine for now, but we should think about auto-generating these docs (/cc @verbanicm).

verbanicm:
@sethvargo Yea we were talking about this offline.

verbanicm (dismissed):

dcreey:
> This is fine for now, but we should think about auto-generating these docs (/cc @verbanicm).

I think it would be nice to have a `--help` formatted as markdown. I think generating the full developer docs from --help might be difficult as we also have sections for pre-reqs and will want to elaborate a lot more in the future on each cli. Right now I just used what is in the guardian --help as a placeholder. But its not very user friendly. If we want good docs we should really add a lot more context/background/details.

verbanicm (approved):

PR #260, opened 2024-01-25
Remove Actions as an Alternative

lock14:
Remove Guardian's github actions as we want users to rely soley on the Guardian CLI for invoking commands.


lock14:
Looks like workflows need to update. @verbanicm should we change the workflows to use equivalent guardian CLI commands, or just remove the workflows? I am assuming we would probably want the former.

verbanicm:
> 

Yea lets update them. We want to make sure we use the binary from the CI run not the "published" one. Not sure what that looks like currently so we might need to build the binary in the CI run and use it instead of the setup-binary action.

verbanicm:
@lock14 There are `abc` cli templates you can run to bootstrap the guardian workflows. Then you can go strip out what isnt needed.

See here: https://github.com/abcxyz/guardian/tree/main/abc.templates/base-workflows

sethvargo (dismissed):

lock14:
Removing `test-apply` workflow as it will break if this PR is merged. We plan to refactor the `test-plan` workflow into a larger workflow that encompasses testing both `plan` and `apply` functionality of guardian at the same time.

verbanicm (approved):

PR #262, opened 2024-01-26
Make tests more comprehensive

### LINKED ISSUE #261: [Tests] Improve integration testing flow
### TL;DR

The current process requires a change to a Terraform resource to generate a change. Additionally, the apply happens after merge, which is ideal in the production workflow but not for testing.

The following changes should be done:

1) Remove all `test-*.yaml` workflows, except `test-plan.yml`
2) Rename `test-plan.yml` to `integration.yml`
3) Update Terraform resources in the `terraform` fodler to have a local backend and "empty" resources. We want to keep multiple folder to ensure the directories command works properly.

> [!NOTE]
> We aren't testing terraform, so these dont need to be "real" resources e.g.
> 
> ```terraform
> terraform {
>   backend "local" {}
> }
> 
> resource "null_resource" "empty" {}
> ```

4) Add an `apply` job after `plan_success` to ensure apply works
5) Add a `destroy` job after `apply` to ensure `guardian run` works

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

lock14:
* Added local terraform backend will null resources to use for testing guardian commands
* renamed `test-plan.yml` -> `integration.yml`
* Added steps to test `apply` and `run` commands for guardians to `integration.yml`
* removed other `test-*.yml` workflows as they are no longer needed.

closes #261


.github/workflows/integration.yml
```diff
@@ -67,17 +65,6 @@ jobs:
           go build -o ${INSTALL_PATH}/guardian cmd/guardian/main.go
           echo "${INSTALL_PATH}" >> $GITHUB_PATH
 
-      - name: 'Guardian Directories'
```
verbanicm, line 70: I think we want to keep this, only remove the detect changes. So the syntax should be:

```
          DIRECTORIES=$(guardian entrypoints -format=json ./.)
          echo "entrypoints -> ${DIRECTORIES}"
          echo "directories=${DIRECTORIES}" >> $GITHUB_OUTPUT
```

verbanicm: This way we "test" the entrypoints also

.github/workflows/integration.yml
```diff
@@ -92,16 +79,14 @@ jobs:
...
-      needs.init.outputs.directories != '[]'
     needs:
       - 'init'
     runs-on: 'ubuntu-latest'
     strategy:
       fail-fast: false
       max-parallel: 100
       matrix:
-        working_directory: '${{ fromJSON(needs.init.outputs.directories) }}'
+        working_directory: ['./terraform/local-backend']
```
verbanicm, line 89: Revert per previous comment about entrypionts

.github/workflows/integration.yml
```diff
@@ -187,3 +171,116 @@ jobs:
...
+    needs:
+      - 'init'
+      - 'plan'
+      - 'plan_success'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: false
+      max-parallel: 100
+      matrix:
+        working_directory: ['./terraform/local-backend']
```
verbanicm, line 185: This should be same as above taking dirs from init.

.github/workflows/integration.yml
```diff
@@ -187,3 +171,116 @@ jobs:
...
+        run: |-
+          guardian apply run \
+          -github-owner="${GITHUB_OWNER_NAME}" \
+          -github-repo="${GITHUB_REPO_NAME}" \
+          -github-token="${REPO_TOKEN}" \
+          -pull-request-number="${PULL_REQUEST_NUMBER}" \
+          -bucket-name="${GUARDIAN_BUCKET_NAME}" \
+          "${DIRECTORY}"
+
+  detroy:
```
verbanicm, line 233: nit: spelling

.github/workflows/integration.yml
```diff
@@ -187,3 +171,116 @@ jobs:
...
+      - 'init'
+      - 'plan'
+      - 'plan_success'
+      - 'apply'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: false
+      max-parallel: 100
+      matrix:
+        working_directory: ['./terraform/local-backend']
```
verbanicm, line 244: Same comment as above about init output dirs

terraform/local-backend/main.tf
```diff
@@ -0,0 +1,5 @@
+terraform {
+  backend "local" {}
+}
+
+resource "null_resource" "empty" {}
```
verbanicm, line 5: nit: newline

.github/workflows/integration.yml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: 'test_guardian_plan'
+name: 'test_guardian'
```
verbanicm, line 15: nit: lets just call this integration

verbanicm (commented):
Can we have two test directories in the terraform folder?

Names dont really matter but something like: 

```
/terraform
  /project1
    main.tf
  /project2
    main.tf
```

.github/workflows/integration.yml
verbanicm, line 233: My mistake, this run command will need to come in the apply job, because the state file wont exist in this job. You can add this as an additional step in the apply job.

.github/workflows/integration.yml
lock14: done.

.github/workflows/integration.yml
lock14: done.

.github/workflows/integration.yml
lock14: done.

.github/workflows/integration.yml
lock14: done.

terraform/local-backend/main.tf
lock14: fixed.

.github/workflows/integration.yml
lock14: done.

.github/workflows/integration.yml
lock14: moved `destroy` command to end of `apply` job.

.github/workflows/integration.yml
lock14: Hmm with this change, directories from `pkg` are not being included

verbanicm: It should be targeted at `./terraform`, not sure why this differs from the template: 

https://github.com/abcxyz/guardian/blob/main/abc.templates/base-workflows/contents/guardian-plan.yml#L75

lock14: Yes, I've updated it

verbanicm (approved):

PR #264, opened 2024-01-29
Add -no-color flag to Terraform Fmt

lock14:
* Add `-no-color` flag to when executing Terraform Fmt to avoid colors in ouput
* Refactor `updateResultCommentForActions()` to reduce redundant code

### DIFF
--- a/pkg/commands/plan/plan_run.go
+++ b/pkg/commands/plan/plan_run.go
@@ -267,7 +267,7 @@ func (c *PlanRunCommand) createStartCommentForActions(ctx context.Context) (*git
 	return startComment, nil
 }
 
-func (c *PlanRunCommand) updateResultCommentForActions(ctx context.Context, startComment *github.IssueComment, result *RunResult, resulErr error) error {
+func (c *PlanRunCommand) updateResultCommentForActions(ctx context.Context, startComment *github.IssueComment, result *RunResult, resultErr error) error {
 	logger := logging.FromContext(ctx)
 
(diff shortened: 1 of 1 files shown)


verbanicm (changes requested):
All commands should support the -no-color flag to prevent this issue, we should be explicitly passing it for the format command, if not this is a bug.

lock14:
> All commands should support the -no-color flag to prevent this issue, we should be explicitly passing it for the format command, if not this is a bug.

We were not providing the `-no-color` flag when calling terrform format, I've updated the code to use that as the solution instead.

pkg/commands/plan/plan_run.go
```diff
@@ -276,55 +276,27 @@ func (c *PlanRunCommand) updateResultCommentForActions(ctx context.Context, star
...
-
-		if err := c.gitHubClient.UpdateIssueComment(
-			ctx,
-			c.GitHubFlags.FlagGitHubOwner,
-			c.GitHubFlags.FlagGitHubRepo,
-			startComment.ID,
-			comment.String(),
-		); err != nil {
-			return fmt.Errorf("failed to update plan error comment: %w", err)
+		if result.hasChanges {
```
verbanicm, line 283: Not sure if this scenarios exits, but just in case, should we invert this? In the case where hasChanges = true and err != nil, we would show success right?

lock14: Good point. I swapped the order so that the `resultErr != nil` case takes precedence.

verbanicm (approved):

PR #268, opened 2024-01-31
fix: update remove comment command to allow for apply comments too

verbanicm:


pkg/commands/workflows/remove_comments.go
```diff
@@ -71,6 +84,13 @@ func (c *RemovePlanCommentsCommand) Flags() *cli.FlagSet {
 		Usage:   "The GitHub pull request number to remove plan comments from.",
 	})
 
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "comment-type",
```
verbanicm, line 88: Not sure about this name/API presentation. I dont want to remove all guardian comments because that is wrong. This allows for different types of comments.

yolocs: Alternative names, e.g. `from-actions` `for-actions`

pkg/util/util.go
```diff
@@ -16,12 +16,23 @@
...
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
+
+	"golang.org/x/exp/maps"
 )
 
+// SortedMapKeys returns the sorted slice of map key strings.
+func SortedMapKeys[M ~map[K]V, K cmp.Ordered, V any](m M) []K {
```
yolocs, line 30: Need to update my Go knowledge. What is `~` used for?

yolocs (dismissed):

pkg/util/util.go
verbanicm: Full transparency, I copied and modified the signature of the maps.Keys function (https://pkg.go.dev/golang.org/x/exp/maps#Keys) and married with the signature of slices.Sort. I believe `~` means, "~T means the set of all types whose underlying type is T" --> https://www.calhoun.io/what-does-the-tilde-mean-in-go-generics/

drevell: 😵‍💫

pkg/commands/workflows/remove_comments.go
```diff
@@ -71,6 +84,13 @@ func (c *RemovePlanCommentsCommand) Flags() *cli.FlagSet {
...
+		Target:  &c.flagCommentTypes,
+		Example: "true",
```
drevell, line 91: `true` doesn't seem right here, maybe a copy-paste oversight

pkg/commands/workflows/remove_comments.go
```diff
@@ -84,13 +104,23 @@ func (c *RemovePlanCommentsCommand) Flags() *cli.FlagSet {
 			merr = errors.Join(merr, fmt.Errorf("missing flag: pull-request-number is required"))
 		}
 
+		if len(c.flagCommentTypes) == 0 {
+			merr = errors.Join(merr, fmt.Errorf("missing flag: comment-type is required"))
+		}
+
+		for _, commentType := range c.flagCommentTypes {
+			if _, ok := commentTypePrefixes[commentType]; !ok {
```
drevell: I think you could use sets.Subtract here, for example https://github.com/abcxyz/abc/blob/3b502883f47d960b9258e566c53c34e52676634f/templates/common/builtinvar/builtin_vars.go#L51

pkg/commands/workflows/remove_comments.go
```diff
@@ -136,12 +167,21 @@ func (c *RemovePlanCommentsCommand) Process(ctx context.Context) error {
 		}
 
 		for _, comment := range response.Comments {
-			if !strings.HasPrefix(comment.Body, plan.CommentPrefix) {
-				continue
-			}
+			for _, commentType := range c.flagCommentTypes {
```
drevell: Triple-nested for loop suggest maybe factoring this out into a standalone function

pkg/commands/workflows/remove_comments.go
verbanicm: Went with `for-command`. Thanks!

drevell (changes requested):

pkg/commands/workflows/remove_comments.go
drevell: Actually,this is fine, never mind

pkg/commands/workflows/remove_comments.go
verbanicm: Gah, good catch, this was from an earlier implementation.

pkg/commands/workflows/remove_comments.go
verbanicm: nice!

drevell (approved):

PR #273, opened 2024-02-02
Make supplying Terraform directory consistent

lock14:
The `run` command was recently changed to supply the Terraform directory via a `-dir` flag. The `plan`, `apply`, and `entrypoints` commands have been similarly updated so that supplying the Terraform directory is consistent across each of guardian's commands.

### DIFF
--- a/pkg/commands/apply/apply.go
+++ b/pkg/commands/apply/apply.go
@@ -66,6 +66,7 @@ type ApplyCommand struct {
 	flags.GitHubFlags
 	flags.RetryFlags
 
+	flagDir                  string
 	flagBucketName           string
 	flagCommitSHA            string
 	flagPullRequestNumber    int
@@ -100,6 +101,13 @@ func (c *ApplyCommand) Flags() *cli.FlagSet {
 	c.RetryFlags.Register(set)
 
 	f := set.NewSection("COMMAND OPTIONS")
+
+	f.StringVar(&cli.StringVar{
+		Name:    "dir",
+		Target:  &c.flagDir,
+		Example: "-dir=terraform",
+		Usage:   "The location of the terraform directory",
+	})
 
 	f.IntVar(&cli.IntVar{
 		Name:    "pull-request-number",
@@ -148,12 +156,11 @@ func (c *ApplyCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	parsedArgs := f.Args()
-	if len(parsedArgs) != 1 {
+	if c.flagDir == "" {
 		return flag.ErrHelp
 	}
 
-	dirAbs, err := util.PathEvalAbs(parsedArgs[0])
+	dirAbs, err := util.PathEvalAbs(c.flagDir)
 	if err != nil {
 		return fmt.Errorf("failed to absolute path for directory: %w", err)
 	}
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -55,6 +55,7 @@ type EntrypointsCommand struct {
 	flags.GitHubFlags
 	flags.RetryFlags
 
+	flagDir                     string
 	flagPullRequestNumber       int
 	flagDestRef                 string
 	flagSourceRef               string
@@ -87,6 +88,13 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 	c.RetryFlags.Register(set)
 
 	f := set.NewSection("COMMAND OPTIONS")
+
+	f.StringVar(&cli.StringVar{
+		Name:    "dir",
+		Target:  &c.flagDir,
+		Example: "-dir=terraform",
+		Usage:   "The location of the terraform directory",
+	})
 
 	f.IntVar(&cli.IntVar{
 		Name:    "pull-request-number",
@@ -165,13 +173,11 @@ func (c *EntrypointsCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	parsedArgs := f.Args()
-
-	if len(parsedArgs) != 1 {
+	if c.flagDir == "" {
 		return flag.ErrHelp
 	}
 
-	dirAbs, err := util.PathEvalAbs(parsedArgs[0])
+	dirAbs, err := util.PathEvalAbs(c.flagDir)
 	if err != nil {
 		return fmt.Errorf("failed to absolute path for directory: %w", err)
 	}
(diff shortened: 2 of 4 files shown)


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -88,6 +89,13 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 
 	f := set.NewSection("COMMAND OPTIONS")
 
+	f.StringVar(&cli.StringVar{
+		Name:    "dir",
+		Target:  &c.flagDir,
+		Example: "-dir=terraform",
```
verbanicm, line 95: Example doesnt usually include the flag name, just the value, throughout.

I believe a previous PR did this do, can you validate throughout?

pkg/commands/apply/apply.go
```diff
@@ -148,12 +156,11 @@ func (c *ApplyCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	parsedArgs := f.Args()
-	if len(parsedArgs) != 1 {
+	if c.flagDir == "" {
```
verbanicm, line 159: This is usually done in an afterparse function and a better error message is given. See other commands.

pkg/commands/entrypoints/entrypoints.go
lock14: fixed.

pkg/commands/apply/apply.go
lock14: fixed.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -88,6 +88,13 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 
 	f := set.NewSection("COMMAND OPTIONS")
 
+	f.StringVar(&cli.StringVar{
+		Name:    "dir",
+		Target:  &c.flagDir,
+		Example: "./terraform",
```
sethvargo, line 94: Should this default to CWD?

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -153,6 +160,10 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 			c.parsedFlagMaxDepth = &c.flagMaxDepth
 		}
 
+		if c.flagDir == "" {
+			merr = errors.Join(merr, fmt.Errorf("missing flag: dir is required"))
```
sethvargo, line 164:
```suggestion
			merr = errors.Join(merr, fmt.Errorf("missing flag: -dir is required"))
```

But see also: should this default to CWD?

pkg/commands/plan/plan.go
```diff
@@ -95,6 +95,13 @@ func (c *PlanCommand) Flags() *cli.FlagSet {
 
 	f := set.NewSection("COMMAND OPTIONS")
 
+	f.StringVar(&cli.StringVar{
```
sethvargo, line 98: Since this is repeated, we should pull it out (like GitHubFlags and RetryFlags)

pkg/commands/plan/plan.go
```diff
@@ -155,12 +166,7 @@ func (c *PlanCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	parsedArgs := f.Args()
```
verbanicm, line 158: I think we still want this, only it should validate there are 0 args otherwise, show help since we dont expect any args.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -165,13 +176,7 @@ func (c *EntrypointsCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	parsedArgs := f.Args()
```
verbanicm, line 168: I think we still want this, only it should validate there are 0 args otherwise, show help since we dont expect any args.

pkg/commands/entrypoints/entrypoints.go
verbanicm: +1, we should make this consistent throughout.

pkg/commands/plan/plan.go
lock14: done.

pkg/flags/common.go
```diff
@@ -0,0 +1,42 @@
...
+	FlagDir string
+}
+
+func (c *CommonFlags) Register(set *cli.FlagSet) {
+	f := set.NewSection("COMMON OPTIONS")
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
```
lock14, line 33: I'm not sure if there's a better way to handle this err.

lock14: I'm not sure that would help as the `Flags()` method does not allow returning an error which is where we would be calling the `Register` method.

https://github.com/abcxyz/guardian/blob/main/pkg/commands/plan/plan.go#L95

pkg/commands/apply/apply.go
```diff
@@ -149,20 +151,24 @@ func (c *ApplyCommand) Run(ctx context.Context, args []string) error {
...
 
-	cwd, err := c.WorkingDir()
+	if c.FlagDir == "" {
+		c.FlagDir = cwd
+	}
+
+	dirAbs, err := util.PathEvalAbs(c.FlagDir)
 	if err != nil {
-		return fmt.Errorf("failed to get current working directory: %w", err)
+		return fmt.Errorf("failed to absolute path for directory: %w", err)
```
verbanicm, line 169: nit: typo

```suggestion
		return fmt.Errorf("failed to get absolute path for directory: %w", err)
```

pkg/commands/apply/apply.go
```diff
@@ -149,20 +151,24 @@ func (c *ApplyCommand) Run(ctx context.Context, args []string) error {
...
-	dirAbs, err := util.PathEvalAbs(parsedArgs[0])
+	cwd, err := c.WorkingDir()
 	if err != nil {
-		return fmt.Errorf("failed to absolute path for directory: %w", err)
+		return fmt.Errorf("failed to get current working directory: %w", err)
 	}
-	c.directory = dirAbs
 
-	cwd, err := c.WorkingDir()
+	if c.FlagDir == "" {
```
verbanicm, line 163: Should this be `c.CommonFlags.FlagDir`? (throughout)

lock14: Due to the way Go handles Struct embedding, `c.FlagDir` and `c.CommonFlags.FlagDir` are equivalent names for the same field. I can change it if you find the other way more readable.

pkg/commands/apply/apply.go
lock14: fixed.

pkg/commands/apply/apply.go
verbanicm: Per offline discussion, we can leave this.

verbanicm (approved):

PR #274, opened 2024-02-05
Document cli GitHub permission commands

lock14:
Specify which GitHub permissions are needed explicitly for each the guardian CLI commands.

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -48,6 +48,10 @@ Determine the entrypoint directories to run Guardian commands.
 
 Usage: guardian entrypoints [options]
 
+### Prerequisites
+
+* Required GitHub [permissions](#guardian-cli).
+
 ### Options
 
 Also supports [GitHub Options](#github-options) and [Retry Options](#retry-options).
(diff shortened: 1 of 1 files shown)


cli.md
```diff
@@ -62,14 +62,14 @@ Also supports [GitHub Options](#github-options) and [Retry Options](#retry-optio
...
+  this plan. The default value is "0".
 * **-source-ref="ref-name"** - The source GitHub ref name for finding file changes.
 
-## Apply Run
+## Apply
 
 Run Terraform apply for a directory.
 
-Usage: guardian apply run [options] <directory>.
+Usage: guardian apply [options] <directory>.
```
verbanicm, line 72: Can you update the `-dir` flag on all commands and remove the <directory> arg?

cli.md
```diff
@@ -2,18 +2,18 @@
...
-| workflows                   | [plan-status-comment](#workflows-plan-status-comment)   | Add Guardian plan comment to a pull request                   |
-|                             | [remove-plan-comments](#workflows-remove-plan-comments) | Remove previous Guardian plan comments from a pull request    |
-|                             | [validate-permissions](#workflows-validate-permissions) | Validate required permissions for the current GitHub workflow |
+| **Command**                 | **Subcommand**                                          | **Required Github Permission**                              | **Description**                                               |
+|-----------------------------|---------------------------------------------------------|-------------------------------------------------------------|---------------------------------------------------------------|
+| [entrypoints](#entrypoints) |                                                         | `contents: read`, `pull-requests: write`                    | Determine the entrypoint directories to run Guardian commands |
```
verbanicm, line 7: nit: I really like this approach, but this would read much better if we could add a newline between each permission needed.....not sure if we can do that with an HTML <br/> or what is supported

verbanicm: Just realized the `<br />` was rendered

cli.md
lock14: done.

cli.md
```diff
@@ -308,7 +312,7 @@ Usage: guardian workflows plan-status-comment [options] <pull_request_number>
 
 ### Prerequisites
 
-* Read permission to the target GitHub repository `pull-requests`.
+* Write permission to the target GitHub repository `pull-requests`.
```
verbanicm, line 315: Feels strange that these two commands are the only ones that call out the github token permissions in the body text, but the rest dont. Ideally we should add the same for all commands or remove these since they are documented up top.

cli.md
lock14: Adding a `<br>` tag appears to do the trick.

cli.md
verbanicm: Or link back to the top when referencing permissions, maintaining in two places might not be ideal.

lock14: went with linking back to the permissions table.

cli.md
```diff
@@ -46,12 +46,13 @@ These options influence how Guardian attempts to retry failed requests:
 
 Determine the entrypoint directories to run Guardian commands.
 
-Usage: guardian entrypoints [options] <directory>
+Usage: guardian entrypoints [options]
```
verbanicm, line 49: For consistency can we add a prerequisites for this command also

lock14: done.

verbanicm (approved):

PR #296, opened 2024-02-21
Remove extra `v` from the download URL

### LINKED ISSUE #294: 'Setup Guardian' step fails with 404
### TL;DR

Running the workflow fails with:

```shell
Error downloading: bad response code: 404
```

This is due to an extra `v` prefix to the version tag in the download URL

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

```markdown
v1.0.1
```

### Additional information

_No response_

gjonathanhong:
Fixes #294

### DIFF
--- a/abc.templates/drift-workflows/spec.yaml
+++ b/abc.templates/drift-workflows/spec.yaml
@@ -69,5 +69,3 @@ steps:
         # re-rendering this template using one of the latest tags at https://github.com/abcxyz/guardian/tags.
         #
         ########################################
-
-
--- a/abc.templates/drift-workflows/spec.yaml
+++ b/abc.templates/drift-workflows/spec.yaml
@@ -54,7 +54,7 @@ steps:
         - to_replace: 'REPLACE_GUARDIAN_IGNORE_DIR_PATTERNS'
           with: '{{.guardian_ignore_dir_patterns}}'
         - to_replace: 'REPLACE_GUARDIAN_VERSION_TAG'
-          with: '{{._git_tag}}'
+          with: '{{ slice ._git_tag 1 }}'
 
   # TODO: once https://github.com/abcxyz/abc/pull/423 is available, switch to a rule requiring a '_git_tag' value
   - desc: 'Print warning if rendered without git tag'
@@ -69,3 +69,5 @@ steps:
         # re-rendering this template using one of the latest tags at https://github.com/abcxyz/guardian/tags.
         #
         ########################################
+
+
--- a/abc.templates/base-workflows/spec.yaml
+++ b/abc.templates/base-workflows/spec.yaml
@@ -58,7 +58,7 @@ steps:
         - to_replace: 'REPLACE_GUARDIAN_STATE_BUCKET'
           with: '{{.guardian_state_bucket}}'
         - to_replace: 'REPLACE_GUARDIAN_VERSION_TAG'
-          with: '{{._git_tag}}'
+          with: '{{ slice ._git_tag 1 }}'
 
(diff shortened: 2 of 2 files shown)


abc.templates/base-workflows/contents/guardian-admin.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: 'Setup Guardian'
         uses: 'abcxyz/pkg/.github/actions/setup-binary@3eaafa17e8e4b078507bed004ad5c57627d0e486' # ratchet:abcxyz/pkg/.github/actions/setup-binary@main
         with:
-          download_url: 'https://github.com/abcxyz/guardian/releases/download/v${{ env.GUARDIAN_VERSION }}/guardian_${{ env.GUARDIAN_VERSION }}_linux_amd64.tar.gz'
+          download_url: 'https://github.com/abcxyz/guardian/releases/download/${{ env.GUARDIAN_VERSION }}/guardian_${{ env.GUARDIAN_VERSION }}_linux_amd64.tar.gz'
```
verbanicm, line 61: I dont think this is gonna work...... here is an example download URL:

The url contains the v, the binary doesnt.

`https://github.com/abcxyz/guardian/releases/download/v1.0.1/guardian_1.0.1_darwin_amd64.tar.gz`

gjonathanhong: Reverted original changes and instead removed the `v` prefix before doing string replacement.

I'm assuming we should only accept valid semver tags, so I didn't check if the `v` prefix exists or not

abc.templates/drift-workflows/spec.yaml
```diff
@@ -54,7 +54,7 @@ steps:
         - to_replace: 'REPLACE_GUARDIAN_IGNORE_DIR_PATTERNS'
           with: '{{.guardian_ignore_dir_patterns}}'
         - to_replace: 'REPLACE_GUARDIAN_VERSION_TAG'
-          with: '{{._git_tag}}'
+          with: '{{ slice ._git_tag 1 }}'
```
lock14, line 57: does this work?

lock14 (changes requested):
https://github.com/abcxyz/guardian/blob/main/abc.templates/base-workflows/spec.yaml should be updated with the same fix as well.

abc.templates/base-workflows/contents/guardian-admin.yml
sethvargo: Our tags will always have a leading "v" because that's what Go expects.

abc.templates/drift-workflows/spec.yaml
lock14: I was able to confirm this does work

abc.templates/base-workflows/contents/guardian-admin.yml
lock14: once the the next version of abc cli is released we can add a rule to check that the git tag matches a semvar format if we want to

sethvargo: I'm also working on updating the release process so this will be codified everywhere

gjonathanhong:
@lock14 thanks for catching, done

abc.templates/drift-workflows/spec.yaml
gjonathanhong: In the future, we can test this using goldentests to provide a test version tag

lock14 (approved):

verbanicm (approved):
I am not a fan of this syntax but....it gets this working

gjonathanhong:
:trollface:

PR #317, opened 2024-04-21
fix: parsing of members attribute with non google provider resources

smaftoul:
When using detect drift, I had this error:
> failed to parse IAM from Terraform State: failed to execute terraform IAM tasks in parallel: failed to parse terraform states: failed to decode terraform state: json: cannot unmarshal object into Go struct field .resources.instances.Attributes.members of type string

It turns out guardian downloaded and parsed a terraform state file including a `github_team_members` resource from the github provider, that has a `members` field, whose value is not an array of string, but an array of object.

Not sure the fix provided is the best way to handle this, but at least, test passes.
We could probably filter resources by provider field, but it would probably mean creating our json decoder.

### DIFF
--- a/pkg/terraform/parser/parser.go
+++ b/pkg/terraform/parser/parser.go
@@ -43,12 +43,12 @@ const (
 // ResourceInstances represents the JSON terraform state IAM instance.
 type ResourceInstance struct {
 	Attributes struct {
-		ID      string   `json:"id"`
-		Members []string `json:"members,omitempty"`
-		Member  string   `json:"member,omitempty"`
-		Folder  string   `json:"folder,omitempty"`
-		Project string   `json:"project,omitempty"`
-		Role    string   `json:"role,omitempty"`
+		ID      string        `json:"id"`
+		Members []interface{} `json:"members,omitempty"`
+		Member  string        `json:"member,omitempty"`
+		Folder  string        `json:"folder,omitempty"`
+		Project string        `json:"project,omitempty"`
+		Role    string        `json:"role,omitempty"`
 	}
 }
 
@@ -191,7 +191,7 @@ func (p *TerraformParser) parseIAMBindingForOrg(ctx context.Context, instances [
 	for _, i := range instances {
 		for _, m := range i.Attributes.Members {
 			iams = append(iams, &assetinventory.AssetIAM{
-				Member:       m,
+				Member:       m.(string),
 				Role:         i.Attributes.Role,
 				ResourceID:   p.OrganizationID,
 				ResourceType: assetinventory.Organization,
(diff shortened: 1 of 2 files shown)


sethvargo:
Hmm - this shouldn't happen, since Go will coerce the object into a string if it can. Do you have an example of the JSON?

smaftoul:
@sethvargo, yes, it's provided in the PR as a test. Removing the changes in the PR but keeping the testdata gives this:

```
--- FAIL: TestDrift_DetectDrift (0.00s)
    --- FAIL: TestDrift_DetectDrift/success_no_drift (0.00s)
        drift_test.go:189: DetectDrift() returned error: failed to parse IAM from Terraform State: failed to execute terraform IAM tasks in parallel: failed to parse terraform states: failed to decode terraform state: json: cannot unmarshal object into Go struct field .resources.instances.Attributes.members of type string
        drift_test.go:192: DetectDrift() returned diff (-want +got):
              (*drift.IAMDrift)(
            -   &{ClickOpsChanges: []string{}, MissingTerraformChanges: []string{}},
            +   nil,
              )
    --- FAIL: TestDrift_DetectDrift/success_all_missing_terraform_drift (0.00s)
        drift_test.go:189: DetectDrift() returned error: failed to parse IAM from Terraform State: failed to execute terraform IAM tasks in parallel: failed to parse terraform states: failed to decode terraform state: json: cannot unmarshal object into Go struct field .resources.instances.Attributes.members of type string
        drift_test.go:192: DetectDrift() returned diff (-want +got):
              (*drift.IAMDrift)(
            -   &{
            -           ClickOpsChanges: []string{},
            -           MissingTerraformChanges: []string{
            -                   "/organizations/1231231/folders/123123123123/roles/viewer/group:m"...,
            -                   "/organizations/1231231/projects/my-project/roles/compute.admin/s"...,
            -                   "/organizations/1231231/roles/browser/group:my-group@google.com",
            -                   "/organizations/1231231/roles/browser/serviceAccount:my-service-a"...,
            -                   "/organizations/1231231/roles/browser/user:dcreey@google.com",
            -           },
            -   },
            +   nil,
              )
FAIL
```

sethvargo:
I think the real problem here is that we shouldn't be parsing that resource type for drift /cc @dcreey

verbanicm:
@sethvargo /cc @smaftoul 

I think you are right Seth, I believe the problem is [here](https://github.com/abcxyz/guardian/blob/main/pkg/terraform/parser/parser.go#L159), where we try to parse the entire statefile, using a specific set of resource instance attributes.

Instead, we might want to generalize the `resources[].instances[]` into an `[]interface{}`, then [here](https://github.com/abcxyz/guardian/blob/main/pkg/terraform/parser/parser.go#L167), when we loop over the state resources, we should validate we have an expected resource and then coerce/parse it into the expected struct we expect.

verbanicm:
Took a stab at this, see https://github.com/abcxyz/guardian/pull/318

smaftoul:
Wow, wanted to have a look at this later in the day ;)

smaftoul:
closing this one in favor of #318 then !

verbanicm:
@smaftoul Appreciate the fix and calling out this bug!

PR #318, opened 2024-04-22
fix: drift detection parsing issue

verbanicm:
This fixes an issue where parsing the Terraform statefile would error during drift detection. See #317 for more details.

### DIFF
--- a/pkg/terraform/parser/parser.go
+++ b/pkg/terraform/parser/parser.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"slices"
 	"strconv"
 	"strings"
 
@@ -56,8 +55,8 @@ type ResourceInstance struct {
 // TerraformState represents the JSON terraform state.
 type TerraformState struct {
 	Resources []struct {
-		Type      string        `json:"type"`
-		Instances []interface{} `json:"instances"`
+		Type      string `json:"type"`
+		Instances []any  `json:"instances"`
 	} `json:"resources"`
 }
 
@@ -172,17 +171,17 @@ func (p *TerraformParser) ProcessStates(ctx context.Context, gcsUris []string) (
 func (p *TerraformParser) parseTerraformStateIAM(ctx context.Context, state TerraformState) ([]*assetinventory.AssetIAM, error) {
 	var iams []*assetinventory.AssetIAM
 	for _, r := range state.Resources {
-		targetResources := []string{
-			"google_organization_iam_binding",
-			"google_folder_iam_binding",
-			"google_project_iam_binding",
-			"google_organization_iam_member",
-			"google_folder_iam_member",
-			"google_project_iam_member",
+		targetResources := map[string]struct{}{
+			"google_organization_iam_binding": {},
+			"google_folder_iam_binding":       {},
+			"google_project_iam_binding":      {},
+			"google_organization_iam_member":  {},
+			"google_folder_iam_member":        {},
+			"google_project_iam_member":       {},
 		}
 
 		// short circuit if we dont find a type we want
-		if !slices.Contains(targetResources, r.Type) {
+		if _, ok := targetResources[r.Type]; !ok {
 			continue
 		}
 
@@ -191,7 +190,7 @@ func (p *TerraformParser) parseTerraformStateIAM(ctx context.Context, state Terr
 			return nil, fmt.Errorf("failed to decode terraform state: %w", err)
 		}
 
-		instances := make([]ResourceInstance, len(r.Instances))
+		instances := make([]*ResourceInstance, len(r.Instances))
 		if err := json.Unmarshal(raw, &instances); err != nil {
 			return nil, fmt.Errorf("failed to decode terraform state: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/terraform/parser/parser.go
```diff
@@ -49,14 +50,14 @@ type ResourceInstance struct {
...
+	} `json:"attributes"`
 }
 
 // TerraformState represents the JSON terraform state.
 type TerraformState struct {
 	Resources []struct {
-		Type      string             `json:"type"`
-		Instances []ResourceInstance `json:"instances"`
+		Type      string        `json:"type"`
+		Instances []interface{} `json:"instances"`
```
sethvargo, line 60: Nit: prefer `any`

pkg/terraform/parser/parser.go
```diff
@@ -159,31 +160,59 @@ func (p *TerraformParser) ProcessStates(ctx context.Context, gcsUris []string) (
...
+			"google_project_iam_binding",
+			"google_organization_iam_member",
+			"google_folder_iam_member",
+			"google_project_iam_member",
+		}
+
+		// short circuit if we dont find a type we want
+		if !slices.Contains(targetResources, r.Type) {
+			continue
+		}
```
sethvargo, line 187: Please use a map instead for O(1) lookup

pkg/terraform/parser/parser.go
```diff
@@ -159,31 +160,59 @@ func (p *TerraformParser) ProcessStates(ctx context.Context, gcsUris []string) (
...
+
+		raw, err := json.Marshal(r.Instances)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode terraform state: %w", err)
+		}
+
+		instances := make([]ResourceInstance, len(r.Instances))
+		if err := json.Unmarshal(raw, &instances); err != nil {
+			return nil, fmt.Errorf("failed to decode terraform state: %w", err)
+		}
```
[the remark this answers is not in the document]
sethvargo: What are you trying to do?

pkg/terraform/parser/parser.go
```diff
@@ -159,31 +160,59 @@ func (p *TerraformParser) ProcessStates(ctx context.Context, gcsUris []string) (
...
+		if !slices.Contains(targetResources, r.Type) {
+			continue
+		}
+
+		raw, err := json.Marshal(r.Instances)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode terraform state: %w", err)
+		}
+
+		instances := make([]ResourceInstance, len(r.Instances))
```
sethvargo, line 194:
```suggestion
		instances := make([]*ResourceInstance, len(r.Instances))
```

pkg/terraform/parser/parser.go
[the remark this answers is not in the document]
verbanicm: Were parsing into a slice of any (https://github.com/abcxyz/guardian/pull/318/files#diff-06c48a637212ad3944309b75d80941c60f39c29fc99da7d27332e59aa0e580faR59).

I want to now type cast that slice into []ResourcesInstances (because I have determine these are the resources I care out and have the expected attributes). I dont think you can cast, so I am marshalling and unmarshalling to parse into the expected struct type.

smaftoul:
For the fun, I did a small experimentation:

```
package main

import (
        "fmt"
        "os"

        "github.com/hashicorp/terraform/states/statefile"
)

func main() {
        f, err := os.Open(os.Args[1])
        if err != nil {
                fmt.Println(err)
        }
        state, err := statefile.Read(f)
        if err != nil {
                fmt.Println(err)
        }

        for moduleName, module := range state.State.Modules {
                for _, stateRes := range module.Resources {
                        targetResources := map[string]struct{}{
                                "google_organization_iam_binding": {},
                                "google_folder_iam_binding":       {},
                                "google_project_iam_binding":      {},
                                "google_organization_iam_member":  {},
                                "google_folder_iam_member":        {},
                                "google_project_iam_member":       {},
                        }
                        if _, ok := targetResources[stateRes.Addr.Resource.Type]; ok {
                                fmt.Println(moduleName, stateRes.Addr.Resource.Type, stateRes.Addr.Resource.Name)
                        }
                }
        }
}
```

This seems to yield the resources we are interested in.
The thing is , this package has moved to `github.com/hashicorp/terraform/internal`, but I was able to `go get` it @v0.15.3 , which I think is the last release before 1.0.

pkg/terraform/parser/parser.go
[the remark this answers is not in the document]
verbanicm: @sethvargo figured it out, if we use json.RawMessage (equivalent to []byte) we can just unmarshall into the expected struct.

verbanicm:
@smaftoul I would prefer not to take a dependency on an "outdated" Terraform package, I think my latest push should solve this issue.

pkg/terraform/parser/parser.go
```diff
@@ -159,34 +165,60 @@ func (p *TerraformParser) ProcessStates(ctx context.Context, gcsUris []string) (
...
 	var iams []*assetinventory.AssetIAM
 	for _, r := range state.Resources {
+		targetResources := map[string]struct{}{
+			"google_organization_iam_binding": {},
+			"google_folder_iam_binding":       {},
+			"google_project_iam_binding":      {},
+			"google_organization_iam_member":  {},
+			"google_folder_iam_member":        {},
+			"google_project_iam_member":       {},
+		}
```
sethvargo, line 187: Nit: sort

sethvargo (approved):

PR #322, opened 2024-05-31
feat: add command for cleaning up empty statefiles

dcreey:
Based on design [here](https://docs.google.com/document/d/117fA872HumSa4WtdsB7qgbtHsB5fNzMA1BvZCdQmiKM/edit?tab=t.0).

This command determines all entrypoints that have changed and then checks to see if the statefiles for those entrypoints are empty. If they are, we delete the entrypoints.


pkg/commands/cleanup/cleanup.go
```diff
@@ -0,0 +1,228 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
sethvargo, line 1: 4

sethvargo (approved):

verbanicm (approved):

PR #328, opened 2024-06-13
refactor: only delete target directory with cleanup

dcreey:
We will run this as a step in the plan job:

```
- name: 'Cleanup'
  if: always()
  run: |-
    guardian cleanup -dir="${{ matrix.working_directory }}"
```


verbanicm (approved):
TODO: add documentation for this command, also around soft delete for gcs buckets

PR #332, opened 2024-06-13
Use CAI to fetch IAM policies in IAM detect-drift

gjonathanhong:
This change leverages Cloud Asset Inventory API to reduce the number of req/min. The previous behavior would invoke the resourcemanager API once per project and folder in the organization but now it will call CAI based on the number of IAM resources / page_size.

Note, `asset.SearchAllIamPolicies` will also include IAM policies of projects and folders in `DELETE_REQUESTED` state, so in a follow up, we should call CAI to fetch all the projects in `DELETE_REQUESTED` state and exclude them using the query for `asset.SearchAllIamPolicies`.

### DIFF
--- a/pkg/assetinventory/assetinventory_mock.go
+++ b/pkg/assetinventory/assetinventory_mock.go
@@ -35,7 +35,7 @@ type MockAssetInventoryClient struct {
 	AssetErr         error
 }
 
-func (m *MockAssetInventoryClient) IAM(ctx context.Context, scope, query string) ([]*AssetIAM, error) {
+func (m *MockAssetInventoryClient) IAM(ctx context.Context, scope, query string, assetTypes []string) ([]*AssetIAM, error) {
 	if m.IAMErr != nil {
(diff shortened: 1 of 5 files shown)


pkg/assetinventory/assetinventory_mock.go
```diff
@@ -35,7 +35,7 @@ type MockAssetInventoryClient struct {
 	AssetErr         error
 }
 
-func (m *MockAssetInventoryClient) IAM(ctx context.Context, scope, query string) ([]*AssetIAM, error) {
+func (m *MockAssetInventoryClient) IAM(ctx context.Context, opts IAMOptions) ([]*AssetIAM, error) {
```
sethvargo, line 38:
```suggestion
func (m *MockAssetInventoryClient) IAM(ctx context.Context, opts *IAMOptions) ([]*AssetIAM, error) {
```

sethvargo (approved):

sailorlqh (approved):

PR #335, opened 2024-06-15
fix: enable quotations in input commands of guardian-admin.yml

### LINKED ISSUE #321: Quoted arugments on Guardian Admin workflows not being properly un/escaped.
### TL;DR

I tried running `apply -replace="random_id.github_abc_updater" -input=false -auto-approve` using a guardian admin workflow, but got 
```
  ╷
  │ Error: Invalid force-replace address "\"random_id.github_abc_updater\""
  │ 
  │ Must begin with a variable name.
  ╵
```

It worked correctly after removing the quotes.

According to hashicorp documentation, quotes should work. https://developer.hashicorp.com/terraform/cli/state/taint

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

Team's infra repo.

### Additional information

_No response_

gjonathanhong:
Fixes #321.

### DIFF
--- a/abc.templates/base-workflows/contents/guardian-admin.yml
+++ b/abc.templates/base-workflows/contents/guardian-admin.yml
@@ -133,4 +133,4 @@ jobs:
         run: |-
           guardian run \
           -dir="${DIRECTORY}" \
-          -- ${TERRAFORM_COMMAND}
+          -- ${TERRAFORM_COMMAND//\"}
--- a/abc.templates/base-workflows/contents/guardian-admin.yml
+++ b/abc.templates/base-workflows/contents/guardian-admin.yml
@@ -127,10 +127,9 @@ jobs:
 
       - name: 'Guardian Admin'
         env:
-          TERRAFORM_COMMAND: '${{ inputs.command }}'
           DIRECTORY: '${{ matrix.working_directory }}'
         shell: 'bash'
         run: |-
           guardian run \
           -dir="${DIRECTORY}" \
-          -- ${TERRAFORM_COMMAND//\"}
+          -- ${{ inputs.command }}


gjonathanhong:
Open to alternatives 👋

abc.templates/base-workflows/contents/guardian-admin.yml
```diff
@@ -133,4 +133,4 @@ jobs:
         run: |-
           guardian run \
           -dir="${DIRECTORY}" \
-          -- ${TERRAFORM_COMMAND}
+          -- ${TERRAFORM_COMMAND//\"}
```
pdewilde, line 136: What is going on here? This looks like you are adding a quote, not dropping.

gjonathanhong: In bash, you can use pattern matching to substitute characters with the format `${parameter//pattern/string}`.

From `man bash`, 
> If there are two slashes separating _parameter_ and _pattern_, all matches of _pattern_ are replaced with _string_. If _string_ is null, matches of pattern are deleted and the ‘/’ following _pattern_ may be omitted.

So in this case, I'm matching all double quotes in the `TERRAFORM_COMMAND` string variable and removing it.

pdewilde: TIL.

Might make sense to add a comment somewhere

pdewilde (approved):

abc.templates/base-workflows/contents/guardian-admin.yml
sethvargo: This is going to remove _all_ quotes, which could be unintended if my `TERRAFORM_COMMAND` is something like `terraform "foo bar"` where I intend "foo bar" to be a single argument. I think you want something like https://stackoverflow.com/questions/9733338/shell-script-remove-first-and-last-quote-from-a-variable

gjonathanhong: Fair argument, but that also doesn't fix the initial problem.

It turns out, saving `inputs.command` into an environment variable `$TERRAFORM_COMMAND` messed with the way the arguments are parsed (I'm assuming it uses the default `IFS` delimiter to parse the variable).  
For example, an input of `plan "foo bar"` will parse to `["plan", "\"foo", "bar\""]`

But using the `inputs.command` directly in GitHub Actions expression syntax produced the desired result.

pdewilde (approved):

PR #349, opened 2024-08-01
feat: add destroy functionality

verbanicm:
Adding the ability to detect removed directories and if a user supplied meta comment of `GUARDIAN_DESTROY=terraform/backends/project1` exists, the supplied directory will be planned and applied as a destroy instead.

This functionality only cover the `entrypoints` command to identify the appropriate directories. The output of this command is a breaking change, as the JSON output has additional fields to support the modified and destroyed directories.

The follow up to this will be to implement the functionality in the plan and apply commands to us the destroy flag.

NOTE: Some of the test data files just had a license header added and can be ignored.

GUARDIAN_DESTROY=terraform/project2

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -17,6 +17,7 @@ name: 'integration'
 on:
   pull_request:
     types:
+      - 'edited'
       - 'opened'
       - 'synchronize'
       - 'reopened'
@@ -134,7 +135,7 @@ jobs:
       fail-fast: false
       max-parallel: 100
       matrix:
-        working_directory: '${{ fromJSON(needs.init.outputs.directories) }}'
+        working_directory: '${{ fromJSON(needs.init.outputs.directories).entrypoints }}'
     steps:
       - name: 'Checkout'
         uses: 'actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332' # ratchet:actions/checkout@v4
@@ -239,7 +240,7 @@ jobs:
       fail-fast: false
       max-parallel: 100
       matrix:
-        working_directory: '${{ fromJSON(needs.init.outputs.directories) }}'
+        working_directory: '${{ fromJSON(needs.init.outputs.directories).entrypoints }}'
     steps:
       - name: 'Checkout'
         uses: 'actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332' # ratchet:actions/checkout@v4
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -21,31 +21,32 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/fs"
+	"os"
+	"path"
+	"slices"
 	"sort"
-	"strings"
 
-	"github.com/posener/complete/v2"
 	"golang.org/x/exp/maps"
 
 	"github.com/abcxyz/guardian/pkg/flags"
 	"github.com/abcxyz/guardian/pkg/git"
+	"github.com/abcxyz/guardian/pkg/modifiers"
 	"github.com/abcxyz/guardian/pkg/terraform"
 	"github.com/abcxyz/guardian/pkg/util"
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/sets"
 )
 
 var _ cli.Command = (*EntrypointsCommand)(nil)
 
-// allowedFormats are the allowed format flags for this command.
-var allowedFormats = map[string]struct{}{
-	"json": {},
-	"text": {},
+// EntrypointsResult is the entryponts command result.
+type EntrypointsResult struct {
+	Entrypoints []string `json:"entrypoints"`
+	Modified    []string `json:"modified"`
+	Destroy     []string `json:"destroy"`
 }
-
-// allowedFormatNames are the sorted allowed format names for the format flag.
-// This is used for printing messages and prediction.
-var allowedFormatNames = util.SortedMapKeys(allowedFormats)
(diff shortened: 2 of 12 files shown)


.github/workflows/integration.yml
verbanicm, line 1: This is for backwards compatibility, workflows will change for new destroy functionality after implemented.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -186,32 +198,84 @@ func (c *EntrypointsCommand) Run(ctx context.Context, args []string) error {
...
+		return fmt.Errorf("failed to find entrypoint directories: %w", err)
+	}
+
+	metaValues := modifiers.ParseBodyMetaValues(ctx, c.FlagBodyContents)
+	logger.DebugContext(ctx, "parsed body meta values", "values", metaValues)
+
+	allEntrypointDirs := make([]string, 0, len(entrypointDirs)+len(removedDirs))
+	allEntrypointDirs = append(allEntrypointDirs, entrypointDirs...)
+	allEntrypointDirs = append(allEntrypointDirs, removedDirs...)
+
```
drevell, line 217: This can be simplified a bit:

`allEntryPointDirs := slices.Concat(nil, entryPointDirs, removedDirs)`

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -233,41 +297,92 @@ func (c *EntrypointsCommand) Process(ctx context.Context) error {
...
-// writeOutput writes the command output.
-func (c *EntrypointsCommand) writeOutput(dirs []string) error {
-	cwd, err := c.WorkingDir()
-	if err != nil {
-		return fmt.Errorf("failed to get current working directory: %w", err)
+// processDestroyMetaValues processes the user supplied meta values for directories to destroy.
+func (c *EntrypointsCommand) processDestroyMetaValues(cwd string, dirs []string, metaValues modifiers.MetaValues) ([]string, error) {
+	destroyDirs := make([]string, 0)
+
+	metaDestroyDirs, ok := metaValues["GUARDIAN_DESTROY"]
```
drevell, line 331: Maybe make `"GUARDIAN_DESTROY"` a top-level const, since it will probably be referenced elsewhere

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -233,41 +297,92 @@ func (c *EntrypointsCommand) Process(ctx context.Context) error {
...
+
+	computedDestroyDirs := make(map[string]struct{}, 0)
+	for _, dir := range metaDestroyDirs {
+		computedDestroyDirs[path.Join(cwd, dir)] = struct{}{}
+	}
+
+	for _, dir := range dirs {
+		if _, ok := computedDestroyDirs[dir]; ok {
+			destroyDirs = append(destroyDirs, dir)
+		}
```
drevell, line 344: I suspect this function could be simplified using https://pkg.go.dev/github.com/abcxyz/pkg/sets#Intersect

pkg/modifiers/modifiers.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Package modifiers provides the functionality to parse meta modifiers from text.
+package modifiers
+
+import (
+	"context"
+	"regexp"
+)
+
+var metaValuesKVRegex = regexp.MustCompile(`(?m)^(GUARDIAN_[A-Z0-9_]+)=(.*)$`)
```
drevell, line 23: I'm not sure that multiline mode makes sense here, can you help me understand why? Referring to `(?m)`

It might be cleaner to iterate over the lines of `contents` and match each line individually. That avoids the risk of having a single "match" that unintentionally spans multiple lines.

pkg/modifiers/modifiers.go
```diff
@@ -0,0 +1,39 @@
...
+
+type MetaValues map[string][]string
+
+func ParseBodyMetaValues(ctx context.Context, contents string) MetaValues {
+	metaValues := make(MetaValues, 0)
+
+	matches := metaValuesKVRegex.FindAllStringSubmatch(contents, -1)
+	for _, v := range matches {
+		if _, ok := metaValues[v[1]]; !ok {
```
drevell, line 32: It might help this be more readable if we assign `v[1]` to a variable name, like `indexOfFoo := v[1]`. Just to make this more clear for anyone who hasn't memorized the meaning of the return values from FindAllStringSubmatch 😀

pkg/modifiers/modifiers.go
```diff
@@ -0,0 +1,39 @@
...
+			metaValues[v[1]] = make([]string, 0)
```
drevell, line 33: I suspect you can leave out this if block. It's legal to append to an empty entry in a map. I.e. it's legal to do

```
metaValues["nonexistentKey"] = append(metaValues["nonexistentKey"], "foo")
```

drevell (approved):

pkg/modifiers/modifiers.go
verbanicm: TIL, thanks!

pkg/modifiers/modifiers.go
verbanicm: Ah, fair, will update.

pkg/commands/entrypoints/entrypoints.go
verbanicm: I suspect, you're right. Done.

pkg/modifiers/modifiers.go
verbanicm: Mmmmm this is fair, updated.

pkg/commands/entrypoints/entrypoints.go
verbanicm: Perfect, thanks!

Issue #179, opened 2023-08-15
Improve action output by using workflow step summaries


verbanicm:
### TL;DR

Use GitHub [step summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) to improve readability and easier viewing of outputs from guardian when run in github actions.

### Detailed design

Using step summaries, add a step summary to show the terraform command outputs (plan, apply, etc) in a streamlined format so users can see information easier in the GitHub UI, rather than digging through the output logs.

PR #351, opened 2024-08-01
Link directly to logs for job in Guardian github comments

Celdir:
Context: https://github.com/abcxyz/guardian/issues/179

Make Guardian github comment log links go directly to the logs for the relevant job instead of to the summary page when possible.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -124,6 +124,7 @@ jobs:
             -for-command="apply"
 
   plan:
+    name: 'plan (${{ matrix.working_directory }})'
     needs:
       - 'init'
     runs-on: 'ubuntu-latest'
@@ -229,6 +230,7 @@ jobs:
           -plan-result="${{ needs.plan.result }}"
 
   apply:
+    name: 'apply (${{ matrix.working_directory }})'
     needs:
       - 'init'
       - 'plan'
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -176,6 +176,7 @@ jobs:
           DIRECTORY: '${{ matrix.working_directory }}'
           # used to create comments on pull requests (access to only this repo)
           REPO_TOKEN: '${{ github.token }}'
+          JOB_NAME: '${{ github.job }} (${{ matrix.working_directory }})'
         run: |-
           guardian plan \
           -github-owner="${GITHUB_OWNER_NAME}" \
@@ -183,7 +184,8 @@ jobs:
           -github-token="${REPO_TOKEN}" \
           -pull-request-number="${PULL_REQUEST_NUMBER}" \
           -bucket-name="${GUARDIAN_BUCKET_NAME}" \
-          -dir="${DIRECTORY}"
+          -dir="${DIRECTORY}" \
+          -job-name="${JOB_NAME}"
 
   plan_success:
     needs:
@@ -280,6 +282,7 @@ jobs:
           COMMIT_SHA: '${{ github.sha }}'
           # used to create comments on pull requests (access to only this repo)
           REPO_TOKEN: '${{ github.token }}'
+          JOB_NAME: '${{ github.job }} (${{ matrix.working_directory }})'
         run: |-
           guardian apply \
           -github-owner="${GITHUB_OWNER_NAME}" \
@@ -287,7 +290,8 @@ jobs:
           -github-token="${REPO_TOKEN}" \
           -pull-request-number="${PULL_REQUEST_NUMBER}" \
           -bucket-name="${GUARDIAN_BUCKET_NAME}" \
-          -dir="${DIRECTORY}"
+          -dir="${DIRECTORY}" \
+          -job-name="${JOB_NAME}"
 
       - name: 'Guardian Destroy'
         env:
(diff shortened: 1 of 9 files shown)


.github/workflows/integration.yml
```diff
@@ -175,14 +175,16 @@ jobs:
           DIRECTORY: '${{ matrix.working_directory }}'
           # used to create comments on pull requests (access to only this repo)
           REPO_TOKEN: '${{ github.token }}'
+          JOB_NAME: '${{ github.job }} (${{ matrix.working_directory }})'
```
verbanicm, line 178: I would also add a `name: 'plan (${{ matrix.working_directory }})'` to the job also. Above line 126 (for both apply and plan workflows)

pkg/commands/apply/apply.go
```diff
@@ -319,6 +327,32 @@ func (c *ApplyCommand) Process(ctx context.Context) (merr error) {
 	return merr
 }
 
+func (c *ApplyCommand) resolveGitHubLogURL(ctx context.Context) string {
```
verbanicm, line 330: I wonder if we should create a helper function on the github client to generate this URL

verbanicm: Ideally generate the URL and we can format the link in the caller

drevell: SGTM, some de-duping of the copy-pasted code seems beneficial.

pkg/commands/apply/apply.go
```diff
@@ -319,6 +327,32 @@ func (c *ApplyCommand) Process(ctx context.Context) (merr error) {
...
+	if !c.FlagIsGitHubActions {
+		logger.DebugContext(ctx, "skipping github log url resolution", "is_github_action", c.FlagIsGitHubActions)
+		return logURL
+	}
+
+	// Link to specific job's logs directly if possible
+	jobs, err := c.gitHubClient.ListJobsForWorkflowRun(ctx, c.GitHubFlags.FlagGitHubOwner, c.GitHubFlags.FlagGitHubRepo, c.cfg.RunID, nil)
+	if err != nil {
+		logger.DebugContext(ctx, "failed to list jobs for workflow run", "err", err)
+	} else {
```
drevell, line 345: By adding a `return` inside the `if err != nil` block, we can remove the `else` and save a level of indentation.

pkg/commands/apply/apply.go
```diff
@@ -319,6 +327,32 @@ func (c *ApplyCommand) Process(ctx context.Context) (merr error) {
...
+		for _, job := range jobs.Jobs {
+			if c.flagJobName == job.Name {
+				logURL = fmt.Sprintf("[[logs](%s/%s/%s/actions/runs/%d/job/%d)]", c.cfg.ServerURL, c.GitHubFlags.FlagGitHubOwner, c.GitHubFlags.FlagGitHubRepo, c.cfg.RunID, job.ID)
```
drevell, line 348: Maybe change this to a `return` to simplify a bit? Or do we specifically need the last job with the given name (in which case please add a code comment explaining how/why there are multiple matching jobs)?

drevell (changes requested):

pkg/github/github.go
```diff
@@ -476,6 +492,51 @@ func (g *GitHubClient) RepoUserPermissionLevel(ctx context.Context, owner, repo,
...
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list jobs for workflow run attempt: %w", err)
+	}
+
+	return &JobsResponse{Jobs: jobs, Pagination: pagination}, nil
+}
+
+func (g *GitHubClient) ResolveJobLogsURL(ctx context.Context, jobName, serverURL, owner, repo string, runID int64) (string, error) {
+	jobs, err := g.ListJobsForWorkflowRun(ctx, owner, repo, runID, nil)
```
verbanicm, line 527: Looks like this response returns an `html_url` field, maybe we should use that instead of generating it ourselves?

https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run

Celdir: Oops, good catch! Done!

pkg/commands/apply/apply.go
Celdir: Refactored this whole bit of logic into the github client and used early return there.

pkg/commands/apply/apply.go
Celdir: Done

pkg/commands/apply/apply.go
Celdir: Done

.github/workflows/integration.yml
Celdir: Done

verbanicm (approved):

pkg/github/github.go
```diff
@@ -81,6 +81,17 @@ type PullRequestResponse struct {
 	Pagination   *Pagination
 }
 
+type Job struct {
```
drevell, line 84: Per https://google.github.io/styleguide/go/decisions#doc-comments, `All top-level exported names must have doc comments...`

drevell (approved):

pkg/github/github.go
Celdir: Added those for `Job` and `JobsResponse`

PR #352, opened 2024-08-01
feat: complete destroy

verbanicm:
Open to suggestions on the output formatting. The integration workflow is not representative of what the end state will look like in production.

Actual flow:

```
init
  -> modified or "just plan" directories
  -> destroy or plan -destroy directories

plan
  business as usual for modified dirs

plan-destroy
  checkout main
  guardian plan with -destroy flag


PR MERGED

init
  -> modified or "just plan" directories
  -> destroy or plan -destroy directories

apply
  business as usual for modified dirs

apply-destroy
  checkout HEAD-1 (`github.events.pull_request.before`)
  guardian apply (parse destroy from GCS plan file metadata)
```

* Checking out main/head-1 is needed because we potentially deleted the directory or want to delete all existing resources.
* Splitting the jobs allows us to customize per operation (plan/destroy).
* Using GCS object metadata, we can determine if we are doing a destroy during the apply phase for informing the end user as needed, this wont affect the actual apply process, which is driven by the plan file contents.

### DIFF
--- a/pkg/commands/apply/apply.go
+++ b/pkg/commands/apply/apply.go
@@ -30,6 +30,7 @@ import (
 	"github.com/sethvargo/go-githubactions"
 
 	"github.com/abcxyz/guardian/pkg/commands/actions"
+	"github.com/abcxyz/guardian/pkg/commands/plan"
 	"github.com/abcxyz/guardian/pkg/flags"
 	"github.com/abcxyz/guardian/pkg/github"
 	"github.com/abcxyz/guardian/pkg/storage"
@@ -43,6 +44,7 @@ import (
 const (
 	ownerReadWritePerms = 0o600
 	CommentPrefix       = "**`🔱 Guardian 🔱 APPLY`** -"
+	DestroyCommentText  = "\n**`🟧 DESTROY`**"
 )
 
 var _ cli.Command = (*ApplyCommand)(nil)
@@ -64,6 +66,7 @@ type ApplyCommand struct {
 	planFilePath              string
 	gitHubLogURL              string
 	computedPullRequestNumber int
+	isDestroy                 bool
 
 	flags.RetryFlags
 	flags.CommonFlags
@@ -359,12 +362,17 @@ func (c *ApplyCommand) createStartCommentForActions(ctx context.Context) (*githu
 
 	c.Outf("Creating start comment")
 
+	destroyText := ""
+	if c.isDestroy {
+		destroyText = DestroyCommentText
+	}
+
 	startComment, err := c.gitHubClient.CreateIssueComment(
 		ctx,
 		c.GitHubFlags.FlagGitHubOwner,
 		c.GitHubFlags.FlagGitHubRepo,
 		c.computedPullRequestNumber,
-		fmt.Sprintf("%s 🟨 Running for dir: `%s` %s", CommentPrefix, c.childPath, c.gitHubLogURL),
+		fmt.Sprintf("%s 🟨 Running for dir: `%s` %s%s", CommentPrefix, c.childPath, c.gitHubLogURL, destroyText),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create start comment: %w", err)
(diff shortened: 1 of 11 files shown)


pkg/commands/apply/apply.go
```diff
@@ -463,12 +476,17 @@ func (c *ApplyCommand) downloadGuardianPlan(ctx context.Context, path string) (p
...
 
-	exitCode, ok := metadata["plan_exit_code"]
+	exitCode, ok := metadata[plan.MetaKeyExitCode]
 	if !ok {
 		return nil, "", fmt.Errorf("failed to determine plan exit code: %w", err)
 	}
 	planExitCode = exitCode
 
+	planOperation, ok := metadata[plan.MetaKeyOperation]
+	if ok && strings.ToLower(planOperation) == plan.OperationDestroy {
```
drevell, line 486: Maybe https://pkg.go.dev/strings#EqualFold here rather than ToLower

pkg/commands/plan/plan.go
```diff
@@ -313,14 +332,19 @@ func (c *PlanCommand) updateResultCommentForActions(ctx context.Context, startCo
...
+		destroyText = DestroyCommentText
+	}
+
+	msgBody := fmt.Sprintf("%s 🟦 No changes for dir: `%s` %s%s", CommentPrefix, c.childPath, c.gitHubLogURL, destroyText)
 
 	if result.hasChanges || resultErr != nil {
 		var comment strings.Builder
 		if resultErr != nil {
-			fmt.Fprintf(&comment, "%s 🟥 Failed for dir: `%s` %s\n\n<details>\n<summary>Error</summary>\n\n```\n\n%s\n```\n</details>", CommentPrefix, c.childPath, c.gitHubLogURL, resultErr)
+			fmt.Fprintf(&comment, "%s 🟥 Failed for dir: `%s` %s%s\n\n<details>\n<summary>Error</summary>\n\n```\n\n%s\n```\n</details>", CommentPrefix, c.childPath, c.gitHubLogURL, destroyText, resultErr)
```
drevell, line 345: Could this be a helper function like `failureMessage(prefix, dir, childPath, url, destroyText string, err error)`? No strong feelings from me, but I see this duplicated around this codebase.

pkg/util/util.go
```diff
@@ -26,6 +26,16 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+// SliceContainsOnly checks that all values of a slice are a single value.
```
drevell, line 29: Consider adding another sentence like "Returns true when the input is an empty slice."

drevell (approved):

pkg/commands/plan/plan.go
verbanicm: 100%, but this is tech debt, I think we should implement a Reporter interface and have different ways of writing output. Ill create an issue.

pkg/commands/apply/apply.go
verbanicm: ah, nice. thanks!

pkg/util/util.go
verbanicm: Hm, I would expect it to be false, updated code, test, and comment.

pkg/commands/plan/plan.go
verbanicm: Created https://github.com/abcxyz/guardian/issues/355

pkg/commands/workflows/plan_status_comments.go
```diff
@@ -84,7 +85,7 @@ func (c *PlanStatusCommentCommand) Flags() *cli.FlagSet {
 		Usage:   "The Guardian init job result status.",
 	})
 
-	f.StringVar(&cli.StringVar{
+	f.StringSliceVar(&cli.StringSliceVar{
```
sethvargo, line 88: Why this change?

pkg/terraform/plan.go
```diff
@@ -45,6 +46,10 @@ func planArgsFromOptions(opts *PlanOptions) []string {
 		args = append(args, "-compact-warnings")
 	}
 
+	if util.PtrVal(opts.Destroy) {
```
sethvargo, line 49: As a follow-up, we should switch to https://github.com/abcxyz/pkg/blob/main/pointer/pointer.go and drop the util package. Filed https://github.com/abcxyz/guardian/issues/356.

pkg/commands/apply/apply.go
```diff
@@ -358,12 +361,17 @@ func (c *ApplyCommand) createStartCommentForActions(ctx context.Context) (*githu
 
 	c.Outf("Creating start comment")
 
+	destroyText := ""
+	if c.isDestroy {
+		destroyText = DestroyCommentText
+	}
+
```
sethvargo, line 368: I see this same pattern repeated a number of times (thing, if destroy, thing = somethingelse). I've usually found that's a hint that using an interface (or separate command) is actually a better option.

pkg/terraform/plan.go
verbanicm: Ack, will change in a follow up (didnt know that existed)

pkg/commands/workflows/plan_status_comments.go
verbanicm: tl;dr - we will have multiple jobs (plan and plan_destroy) to account for having to checkout different branches (pr sha for plan, main for plan_destroy). This means we will have an end status comment based on multiple job results.

pkg/commands/apply/apply.go
verbanicm: I was planning on creating a reporter interface, so this info can be passed of to create different strategies to create output, ie github, stdout, some other thing. Will resolve this in a follow up.

PR #358, opened 2024-08-06
Aggregate missing team and user approvals from policy evaluation results

gjonathanhong:


pkg/commands/policy/policy_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			c := &PolicyCommand{
+				results: tc.results,
+			}
+
+			err := c.Process(ctx)
+			if diff := testutil.DiffErrString(err, tc.wantErr); diff != "" {
```
gjonathanhong, line 99: question for reviewer: do i need to cuddle the err here if I'm already cuddling diff?

pkg/commands/policy/policy.go
```diff
@@ -90,5 +96,40 @@ func (c *PolicyCommand) Run(ctx context.Context, args []string) error {
...
+
+		if len(v.MissingApprovals) == 0 {
+			logger.DebugContext(ctx, "no missing approvals for policy",
+				"policy_name", k)
+			continue
+		}
+
+		for _, m := range v.MissingApprovals {
+			c.teams = append(c.teams, m.AssignTeams...)
+			c.users = append(c.users, m.AssignUsers...)
```
verbanicm, line 122: This seems unnecessary and is only used to try to get the test results you want. I would just test the response message. When you add the functionality to assign reviewer, mock the GitHub client and expect the requests sent contain these reviewers instead (similiar to how plan and apply are tested).

pkg/commands/policy/policy.go
```diff
@@ -90,5 +96,40 @@ func (c *PolicyCommand) Run(ctx context.Context, args []string) error {
...
 
-	return nil
+	c.results = results
+
+	return c.Process(ctx)
+}
+
+// Process handles the main logic for handling the results of the policy
+// evaluation.
+func (c *PolicyCommand) Process(ctx context.Context) error {
```
verbanicm, line 106: You could accept the results as a param here or move the code process the results file into the process function and use a file from a testdata folder inside this dir to test the functionality fully.

pkg/commands/policy/policy_test.go
verbanicm: I believe this is correct.

pkg/commands/policy/policy.go
gjonathanhong: Removed from struct and moved all logic (except for parsing flags) into `Process`

pkg/commands/policy/policy.go
gjonathanhong: Opted for latter

pkg/commands/policy/testdata/results.json
verbanicm, line 1: Is this file used?

verbanicm (approved):

sethvargo (approved):

pkg/commands/policy/testdata/results.json
gjonathanhong: 💥 💣💥 gone 💥 💣💥

  😁

PR #359, opened 2024-08-08
Add RequestReviewers to GitHub client

gjonathanhong:


pkg/github/github.go
```diff
@@ -540,6 +549,39 @@ func (g *GitHubClient) ResolveJobLogsURL(ctx context.Context, jobName, owner, re
 	return "", fmt.Errorf("failed to resolve direct URL to job logs: no job found matching name %s", jobName)
 }
 
+// RequestReviewers assigns a list of users and teams as reviewers of a target
+// Pull Request. Mixing existing reviewers in pending review state with new
+// reviewers will result in a no-op without errors thrown.
```
gjonathanhong, line 554: > Mixing existing reviewers in pending review state with new reviewers will result in a no-op without errors thrown.

I really wish this wasn't the case 🤢

pkg/github/github.go
```diff
@@ -94,6 +94,11 @@ type JobsResponse struct {
 	Pagination *Pagination
 }
 
+type RequestReviewersResponse struct {
```
verbanicm, line 97: Add doc comment here.

PR #361, opened 2024-08-08
Parse required github params from actions context in Policy command

gjonathanhong:
This change retrieves and validates parameters from the github context, which are required to assign reviewers via the GitHub API.


gjonathanhong:
Note - base is not `main`

pkg/commands/policy/policy.go
```diff
@@ -97,6 +99,16 @@ func (c *PolicyCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to get token: %w", err)
 	}
 
+	gitHubParams := &GitHubParams{}
```
sethvargo, line 102:
```suggestion
	var gitHubParams GitHubParams
```

sethvargo (approved):

verbanicm (approved):

PR #363, opened 2024-08-09
Assign missing reviewers to pull request in policy command

gjonathanhong:


pkg/commands/policy/policy.go
```diff
@@ -163,6 +163,29 @@ func (c *PolicyCommand) Process(ctx context.Context) error {
 		)
 	}
 
-	// TODO: assign principals as reviewers to current pull request
+	// Make a request per user and team to avoid uncaught behavior from the GitHub
+	// API, which does not assign any of the provided reviewers if any of the
+	// principals exist on the pending review list.
+	for _, u := range users {
```
verbanicm, line 169: Given my recent struggles with decoupling github from our implementation, it might be good to find a way to abstract this now rather than later. IE, PolicyEngine interface or somethign that has a RequestReviewer function (if i were to run this locally it would be a no-op, if i say use the github policy engine it adds a reviewer to the PR, etc). We can talk more offline.

pkg/github/github_mock.go
```diff
@@ -255,7 +255,7 @@ func (m *MockGitHubClient) RequestReviewers(ctx context.Context, owner, repo str
 	}
 
 	return &RequestReviewersResponse{
-		Users: m.UserReviewers,
-		Teams: m.TeamReviewers,
+		Users: append(m.UserReviewers, users...),
+		Teams: append(m.TeamReviewers, teams...),
```
verbanicm, line 259: I think you need to lock these before updating them. Regardless, I dont think we want to append the users/teams, either have a static response (m.UserReviewers) or just return the inputs as we dont really care what this is doing, we just want to expect a specific response.

pkg/flags/global.go
```diff
@@ -49,21 +50,24 @@ func (g *GlobalFlags) Register(set *cli.FlagSet) {
...
+		case "local":
 		default:
-			if os.Getenv("GITHUB_ACTIONS") == "true" {
+			if v, _ := strconv.ParseBool(set.GetEnv("GITHUB_ACTIONS")); v {
 				g.FlagPlatform = "github"
 				break
 			}
 
 			g.FlagPlatform = "local"
 		}
```
sethvargo, line 69: I think you can drop all of this now:

```go
if g.FlagPlatform == "" {
	if v, _ := strconv.ParseBool(set.GetEnv("GITHUB_ACTIONS")); v {
		g.FlagPlatform = "github"
	}
}
```

pkg/commands/policy/github.go
verbanicm, line 1: It is strange to have this exist in the policy command. We have a github package, flags/github, and now this. We should try to consolidate, my gut is telling me we keep the flags/github for exposing hiddne flags for github and then have a map context function in that flag package. if the caller wants to map, great and they can validate they have the flags they want in their own logic.

pkg/flags/global.go
gjonathanhong: Moved to https://github.com/abcxyz/guardian/pull/367

PR #365, opened 2024-08-12
Add Platform flag to new GlobalFlags section

gjonathanhong:
This change helps provide context about which CI/CD environment a Guardian command is running in. This is useful for determining which client implementations to initialize.

### DIFF
--- a/pkg/flags/global.go
+++ b/pkg/flags/global.go
@@ -51,6 +51,11 @@ func (g *GlobalFlags) Register(set *cli.FlagSet) {
 
 	set.AfterParse(func(merr error) error {
 		g.FlagPlatform = strings.ToLower(strings.TrimSpace(g.FlagPlatform))
+
+		if _, ok := allowedPlatforms[g.FlagPlatform]; !ok && g.FlagPlatform != "" {
+			merr = errors.Join(merr, fmt.Errorf("unsupported value for platform flag: %s", g.FlagPlatform))
+		}
+
 		switch g.FlagPlatform {
 		case "github":
 		case "local":
@@ -61,10 +66,6 @@ func (g *GlobalFlags) Register(set *cli.FlagSet) {
 			}
 
 			g.FlagPlatform = "local"
-		}
-
-		if _, ok := allowedPlatforms[g.FlagPlatform]; !ok {
-			merr = errors.Join(merr, fmt.Errorf("unsupported value for platform flag: %s", g.FlagPlatform))
 		}
 
 		return merr
--- a/pkg/flags/global.go
+++ b/pkg/flags/global.go
@@ -27,6 +27,8 @@ var allowedPlatforms = map[string]struct{}{
 	"github": {},
 }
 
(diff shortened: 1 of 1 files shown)


pkg/flags/global.go
```diff
@@ -0,0 +1,69 @@
...
+	// 3. Default value of "local"
+	f.StringVar(&cli.StringVar{
+		Name:    "platform",
+		Target:  &g.FlagPlatform,
+		Example: "github",
+		Usage:   "The code review platform for Guardian to integrate with.",
+	})
+
+	set.AfterParse(func(merr error) error {
+		switch g.FlagPlatform {
```
sethvargo, line 52:
```suggestion
		c.FlagPlatform = strings.ToLower(strings.TrimSpace(g.FlagPlatform))
		switch g.FlagPlatform {
```

pkg/flags/global.go
```diff
@@ -0,0 +1,69 @@
...
+		case "github":
+		default:
+			if os.Getenv("GITHUB_ACTIONS") == "true" {
```
sethvargo, line 55:
```suggestion
			if v, _ := strconv.ParseBool(set.GetEnv("GITHUB_ACTIONS")); v {
```

Two things:

1. Use https://github.com/abcxyz/pkg/blob/main/cli/flags.go#L295
2. Parse the value as a boolean (default is false on bad input)

pkg/flags/global.go
```diff
@@ -0,0 +1,69 @@
...
+	f.StringVar(&cli.StringVar{
+		Name:    "platform",
+		Target:  &g.FlagPlatform,
+		Example: "github",
+		Usage:   "The code review platform for Guardian to integrate with.",
+	})
+
+	set.AfterParse(func(merr error) error {
+		switch g.FlagPlatform {
+		case "github":
```
verbanicm, line 53: We should expose local also, so even if we're running in actions we can force local if needed.

pkg/flags/global.go
gjonathanhong: neat! 🤯

pkg/flags/global.go
```diff
@@ -0,0 +1,72 @@
...
+				g.FlagPlatform = "github"
+				break
+			}
+
+			g.FlagPlatform = "local"
+		}
+
+		if _, ok := allowedPlatforms[g.FlagPlatform]; !ok {
+			merr = errors.Join(merr, fmt.Errorf("unsupported value for platform flag: %s", g.FlagPlatform))
+		}
```
verbanicm, line 68: Isn't this kinda useless since it will be defaulted to local in the default block? Im fine either way (default local, fail unknown)

sethvargo (approved):

pkg/flags/global.go
gjonathanhong: Good point, it's not doing anything significant in its current position. From a UX perspective, I think we should fail fast if we provide `--platform=bitbucket`, instead of defaulting to "local".

I moved this check further up to catch this. PTAL

verbanicm (approved):

PR #368, opened 2024-08-14
Add TokenSource function to github package

gjonathanhong:
This is a generalized version of the TokenSource function, which we will need frequently when instantiating new GitHub clients. The previous implementation tied us to generating `TokenSource`s from `flags.GitHubFlags`

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -598,7 +598,9 @@ func (g *GitHubClient) withRetries(ctx context.Context, retryFunc retry.RetryFun
 	return nil
 }
 
-type GitHubAppInputs struct {
+type TokenSourceInputs struct {
+	GitHubToken string
+
 	GitHubAppID             string
 	GitHubAppPrivateKeyPEM  string
 	GitHubAppInstallationID string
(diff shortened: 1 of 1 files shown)


pkg/github/github.go
```diff
@@ -596,3 +597,38 @@ func (g *GitHubClient) withRetries(ctx context.Context, retryFunc retry.RetryFun
 	}
 	return nil
 }
+
+type GitHubAppInputs struct {
```
verbanicm, line 601: TokenSourceInputs?

pkg/github/github.go
```diff
@@ -596,3 +597,40 @@ func (g *GitHubClient) withRetries(ctx context.Context, retryFunc retry.RetryFun
...
+	GitHubAppPrivateKeyPEM  string
+	GitHubAppInstallationID string
+	GitHubRepo              string
+	Permissions             map[string]string
+}
+
+// TokenSource creates a token source from a GitHub token or GitHub App used for
+// authenticating a github client.
+func TokenSource(ctx context.Context, inputs *TokenSourceInputs) (githubauth.TokenSource, error) {
+	if inputs.GitHubToken != "" {
```
verbanicm, line 614: You should check if inputs is nil and throw an error. This will panic otherwise when its nil.

pkg/github/github.go
```diff
@@ -596,3 +597,40 @@ func (g *GitHubClient) withRetries(ctx context.Context, retryFunc retry.RetryFun
...
+		githubTokenSource, err := githubauth.NewStaticTokenSource(inputs.GitHubToken)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create github static token source: %w", err)
+		}
+		return githubTokenSource, nil
+	} else {
```
verbanicm, line 620: Its idiomatic to not have the else here, since you return above in the if, less indentation.

verbanicm (approved):

PR #369, opened 2024-08-14
feat: start implementing reporter

verbanicm:
Getting there.....

```bash
go run cmd/guardian/main.go run -dir=terraform/project1 -- plan -no-color

--------------------
Starting Guardian Run
--------------------
Running Terraform init

--------------------
Initializing Terraform
--------------------
Initializing the backend...
Initializing provider plugins...
- Reusing previous version of hashicorp/null from the dependency lock file
- Installing hashicorp/null v3.2.2...
- Installed hashicorp/null v3.2.2 (signed by HashiCorp)

╷
│ Warning: Provider lock file not updated
│ 
│ Changes to the provider selections were detected, but not saved in the .terraform.lock.hcl file. To record these selections, run "terraform init" without the "-lockfile=readonly" flag.
╵
Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

--------------------
Running Terraform Command
--------------------

Error: Required plugins are not installed

The installed provider plugins are not consistent with the packages selected in the dependency lock file:
  - registry.terraform.io/hashicorp/null: the cached package for registry.terraform.io/hashicorp/null 3.2.2 (in .terraform/providers) does not match any of the checksums recorded in the dependency lock file

Terraform uses external plugins to integrate with a variety of different infrastructure services. To download the plugins required for this configuration, run:
  terraform init
failed to run command: failed to run command: exit status 1
exit status 1
```


pkg/reporter/reporter.go
```diff
@@ -32,15 +37,50 @@ const (
...
+}
+
+type ReporterCommand struct {
+	cli.BaseCommand
+	flags.GitHubFlags
+	FlagReporter string
+}
+
+// ResolveReporterClient resolves and generates the reporter client based on the reporter flag.
+func (r *ReporterCommand) ResolveReporterClient(ctx context.Context) (Reporter, error) {
```
verbanicm, line 64: Is this a bad pattern? I want to make this available to the commands that need it.

verbanicm: Need to re-think this. Need a common way for multiple commands to use a factory pattern to resolve a reporter. But the inputs to each reporter are different, but will come from a common set flags to the command.

pkg/commands/workflows/plan_status_comments.go
```diff
@@ -132,99 +109,68 @@ func (c *PlanStatusCommentCommand) Run(ctx context.Context, args []string) error
...
+	return c.Process(ctx)
+}
 
-	tokenSource, err := c.GitHubFlags.TokenSource(ctx, map[string]string{
-		"pull_requests": "write",
-	})
-	if err != nil {
-		return fmt.Errorf("failed to get token source: %w", err)
+// createReporterClient creates a new reporter client based on the reporter flag.
+func (c *PlanStatusCommentCommand) createReporterClient(ctx context.Context) (reporter.Reporter, error) {
```
verbanicm, line 122: This function has to be duplicated across commands because each reporter type requires a different set of inputs that come from the command. My previous version created an embedded "ReporterCommand" that embeded this function to allow this to be added to any command as needed.

verbanicm:
Have a different idea here...

PR #371, opened 2024-08-15
feat: start implementing reporter

verbanicm:


pkg/commands/workflows/plan_status_comments.go
```diff
@@ -132,95 +109,68 @@ func (c *PlanStatusCommentCommand) Run(ctx context.Context, args []string) error
...
+	c.reporterClient = rc
+
+	return c.Process(ctx)
+}
 
-	c.cfg = &PlanStatusCommentsConfig{}
-	if err := c.cfg.MapGitHubContext(actionsCtx); err != nil {
-		return fmt.Errorf("failed to load github context: %w", err)
+// createReporterClient creates a new reporter client based on the reporter flag.
+func (c *PlanStatusCommentCommand) createReporterClient(ctx context.Context) (reporter.Reporter, error) {
```
verbanicm, line 122: I dont like this, it needs to be repeated everywhere a reporter needs to be created for a command, which is apply, plan, plan_status_comments, remove_comments, entrypoints

pkg/commands/workflows/plan_status_comments.go
```diff
@@ -66,16 +56,16 @@ Usage: {{ COMMAND }} [options]
...
-		Name:    "pull-request-number",
-		Target:  &c.flagPullRequestNumber,
-		Example: "100",
-		Usage:   "The GitHub pull request number to remove plan comments from.",
+	f.StringVar(&cli.StringVar{
+		Name:    "reporter",
+		Target:  &c.flagReporter,
+		Example: "github",
+		Default: "local",
+		Usage:   "The reporting strategy for Guardian status updates.",
```
drevell, line 68: MIght be helpful to enumerate all the possible values if this flag in the Usage string. It looks like there are probably just two, but it's hard to say for sure.

pkg/commands/workflows/plan_status_comments.go
```diff
@@ -132,95 +108,46 @@ func (c *PlanStatusCommentCommand) Run(ctx context.Context, args []string) error
...
-		With("plan_result", c.flagPlanResult)
-
-	logger.DebugContext(ctx, "determining plan status...")
-
-	c.gitHubLogURL = fmt.Sprintf("[[logs](%s/%s/%s/actions/runs/%d/attempts/%d)]", c.cfg.ServerURL, c.GitHubFlags.FlagGitHubOwner, c.GitHubFlags.FlagGitHubRepo, c.cfg.RunID, c.cfg.RunAttempt)
-	logger.DebugContext(ctx, "computed github log url", "github_log_url", c.gitHubLogURL)
-
 	// this case improves user experience, when the planning job does not have a plan diff
 	// there are no comments, this informs the user that the job ran successfully
 	if c.flagInitResult == "success" && util.SliceContainsOnly(c.flagPlanResult, "success") {
```
drevell, line 124: The string "success" is hardcoded twice here, should it become a constant?

pkg/reporter/github_reporter.go
```diff
@@ -204,6 +157,43 @@ func (g *GitHubReporter) CreateStatus(ctx context.Context, st Status, p *Params)
...
+			}
+
+			// found the prefix, delete the comment
+			if err := g.gitHubClient.DeleteIssueComment(ctx, g.inputs.GitHubOwner, g.inputs.GitHubRepo, comment.ID); err != nil {
+				return fmt.Errorf("failed to delete comment: %w", err)
+			}
+		}
+
+		if response.Pagination == nil {
+			break
```
drevell, line 189: Nit: there's an inconsistency here between `break` here at line 189 and `return nil` at line 173. I'd suggest picking one way of breaking from the loop and using it in both places.

pkg/reporter/reporter.go
```diff
@@ -32,15 +37,47 @@ const (
...
+
+	// ClearStatus clears any existing statuses that can be removed.
+	ClearStatus(ctx context.Context) error
+}
+
+type Config struct {
+	GitHub github.Config
+}
+
+func NewReporter(ctx context.Context, t string, c *Config, stdout io.Writer) (Reporter, error) {
```
drevell, line 61: Technically, exported top-level types and functions should be documented 🤷

drevell (approved):

pkg/reporter/github_reporter.go
verbanicm: ah i migrated this and made it simpler, refactoring issue thanks!

pkg/reporter/reporter.go
verbanicm: d'oh!

pkg/commands/workflows/plan_status_comments.go
verbanicm: this is fair, added

pkg/commands/workflows/plan_status_comments.go
verbanicm: done

PR #372, opened 2024-08-15
Add Local and GitHub implementation of ChangeRequest interface

gjonathanhong:
This change defines a new ChangeRequest interface, which must implement an `AssignReviewers()` function to call a platform's (e.g. GitHub, GitLab, etc.) API.


pkg/platform/pullrequest.go
```diff
@@ -0,0 +1,81 @@
...
+	}, nil
+}
+
+// AssignReviewers calls the GitHub API to assign users and teams as reviewers
+// for the current pull request.
+func (p *PullRequest) AssignReviewers(ctx context.Context, inputs *AssignReviewersInput) error {
+	if inputs == nil {
+		return fmt.Errorf("inputs cannot be nil")
+	}
+	if _, err := p.client.RequestReviewers(ctx, p.params.Owner, p.params.Repository, p.params.PullRequestNumber, inputs.Users, inputs.Teams); err != nil {
```
gjonathanhong, line 77: I think GitHub silently fails if one of the reviewers exists on the PR, so I may need to add some validation of the response or indicate partial assignments (or fail completely if none of the reviewers were added).

pkg/platform/pullrequest.go
```diff
@@ -0,0 +1,81 @@
...
+
+// PullRequest is the GitHub implementation for the ChangeRequest interface.
+type PullRequest struct {
+	client github.GitHub
+	params *PullRequestInput
+}
+
+// NewPullRequest creates a new PullRequest instance.
+func NewPullRequest(ctx context.Context, inputs *PullRequestInput) (*PullRequest, error) {
+	tokenSource, err := github.TokenSource(ctx, &github.TokenSourceInputs{
```
sethvargo, line 48: I think this is gonna conflict with some of the inflight changes that @verbanicm has. Can you two sync on an order of operations?

pkg/platform/pullrequest.go
sethvargo: Is that a problem? I feel like we only care about the end result (that the reviewers are assigned)

pkg/platform/pullrequest_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/abcxyz/guardian/pkg/github"
+	"github.com/abcxyz/pkg/testutil"
+)
+
+const (
+	testOwner             = "test-owner"
+	testRepo              = "test-repo"
+	testPullRequestNumber = 1
```
sethvargo, line 31: Nit: I generally prefer these to be hard coded inline, even if it's repetitive.

sethvargo (approved):

pkg/platform/pullrequest.go
gjonathanhong: I would expect that if I called `AssignReviewers()`, then it would only succeed if all the users and teams that I provided were actually assigned. 

However, if I called `AssignReviewers(Mike, Seth, Brad)` and the pull request already had Seth assigned to it, then the result will be a no-op, where Mike and Brad are not assigned.

sethvargo: Oh interesting. I wonder if we need to call RequestReviewers individually?

pkg/platform/pullrequest.go
verbanicm: Lets get this merged first and ill rebase my changes on to this. I started fixing the other commands so this is just one more.

pkg/platform/pullrequest.go
```diff
@@ -0,0 +1,109 @@
...
+		client: client,
+		params: inputs,
+	}, nil
+}
+
+// AssignReviewers calls the GitHub API to assign users and teams as reviewers
+// for the current pull request. GitHub's RequestReviewer API will result in
+// silent no-op when adding an existing reviewer with new reviewers. Instead,
+// we assign every principal individually to avoid the ambiguous API behavior.
+func (p *PullRequest) AssignReviewers(ctx context.Context, inputs *AssignReviewersInput) (*AssignReviewersResult, error) {
```
gjonathanhong, line 76: I changed the return type to `*AssignReviewersResult` so that we can return the subset of users/teams that were successfully assigned and ignore/log any partial errors

pkg/platform/pullrequest.go
gjonathanhong: Revised, please take another look :)

gjonathanhong:
Synced offline with Mike, have new ideas - need to refactor

PR #373, opened 2024-08-16
fix: refactor github client for plan

verbanicm:
This change uses the new reporter interface for writing comments, which had a dependency on the new github config pattern. 

Also minor clean up on some naming.


pkg/flags/global.go
```diff
@@ -23,9 +23,14 @@ import (
 	"github.com/abcxyz/pkg/cli"
 )
 
+const (
+	PlatformTypeLocal  = "local"
```
drevell, line 27: These constants seem to be the same as the similarly-named constants in reporter.go. Do we really need two copies?

drevell (approved):

pkg/flags/global.go
verbanicm: I think this is a coincidence, but are technically different and may diverge.

PR #377, opened 2024-08-16
Add platform interface

gjonathanhong:
This change defines a new interface named `Platform`, which will define the set of functionality that should be supported across different code review platforms. This is mostly setup before implementation, however, the plan is to refactor the `github.GitHub` interface into a more generalized pattern, in a long term effort to decouple from GitHub specific details.


pkg/platform/platform.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var _ Platform = (*github.GitHubClient)(nil)
+
+// Platform defines the minimum interface for a code review platform.
+type Platform interface{}
+
+// Config is the configuration needed to generate different reporter types.
+type Config struct {
+	GitHub github.Config
```
verbanicm, line 39:
```suggestion
	GitHub *github.Config
```

Apparently I missed that in reporter also

pkg/platform/platform.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/abcxyz/guardian/pkg/github"
+)
+
+const (
+	TypeLocal  string = "local"
+	TypeGitHub string = "github"
```
verbanicm, line 29: These exist in the flags/global.go file now too, we should pick a single place (probably here) and refactor.

pkg/platform/local.go
```diff
@@ -0,0 +1,13 @@
+package platform
+
+import "context"
+
+var _ Platform = (*Local)(nil)
+
+// // Local implements the Platform interface for running Guardian locally.
```
verbanicm, line 7:
```suggestion
// Local implements the Platform interface for running Guardian locally.
```

verbanicm (dismissed):

pkg/platform/platform.go
gjonathanhong: Moved to platform package

pkg/platform/platform.go
```diff
@@ -0,0 +1,102 @@
...
+	// Type value is loaded in the following order:
+	//
+	// 1. Explicit value set through --platform flag
+	// 2. Inferred environment from well-known environment variables
+	// 3. Default value of "local"
+	f.StringVar(&cli.StringVar{
+		Name:    "platform",
+		Target:  &c.Type,
+		Example: "github",
+		Usage:   "The code review platform for Guardian to integrate with.",
```
verbanicm, line 64: We should add the allowed values here.

verbanicm (approved):

pkg/platform/platform.go
```diff
@@ -0,0 +1,108 @@
...
+	f := set.NewSection("PLATFORM OPTIONS")
+	// Type value is loaded in the following order:
+	//
+	// 1. Explicit value set through --platform flag
+	// 2. Inferred environment from well-known environment variables
+	// 3. Default value of "local"
+	i := 0
+	types := make([]string, len(allowedTypes))
+	for k := range allowedTypes {
+		types[i] = k
```
verbanicm, line 63: See reporter.go I believe there should be an example there of how to do this.

verbanicm (approved):

PR #381, opened 2024-08-18
fix: create plan storage client

verbanicm:
Refactor the implementation of storage for plan/apply. This removes the logic of creating storage paths specific to the storage implementation from plan/apply and moves it into its on PlanStorage client.

There is a lot of coupling between several components in the cleanup, drift, and parser packages. This is a best effort attempt to keep changes minimal.

STACKED_PR=https://github.com/abcxyz/guardian/pull/377


pkg/storage/plan_storage.go
```diff
@@ -0,0 +1,127 @@
...
+	"strings"
+
+	"github.com/abcxyz/guardian/pkg/github"
+	"github.com/abcxyz/guardian/pkg/platform"
+)
+
+var _ PlanStorage = (*PlanStorageClient)(nil)
+
+// PlanStorageClient implements [PlanStorage].
+type PlanStorageClient struct {
```
sethvargo, line 33: I'm confused by this encapsulation (and the overload of the word "storage"). I feel like the client should be constructed outside and passed in as a parameter instead.

verbanicm: to me the storage flag is "owned" by the plan/apply command. it should be used to parse a storage type and generate a client.

separately we need configs from the platform to generate storage urls (e.g. the path prefix for uniqueness when using github and pull requests).

does it make sense to add a helper function to the platform config to generate the required "plan storage configs", accepting the storage flag value?

sethvargo: > to me the storage flag is "owned" by the plan/apply command

I think that's correct, but why is it generating a specific client for Plan vs Apply? It should be creating a generic "storage" interface (CREATE, READ, DELETE) for gcs and local. Then in the "storage" package, we instantiate the storage client from the flag inputs.

verbanicm: updated per offline discussions

sethvargo (approved):

PR #384, opened 2024-08-20
Implement AssignReviewers to platform interface

gjonathanhong:


pkg/platform/github.go
```diff
@@ -17,62 +17,67 @@ package platform
...
 )
 
 var _ Platform = (*GitHub)(nil)
 
 // GitHub implements the Platform interface.
 type GitHub struct {
-	cfg    *gh.Config
-	client *github.Client
+	cfg    *github.Config
+	client github.GitHub
```
gjonathanhong, line 30: @verbanicm I've changed my approach a bit here. I think we need to keep the `github` package for mocking the GitHub client.

pkg/platform/github.go
```diff
@@ -17,62 +17,67 @@ package platform
...
+	return &GitHub{
+		cfg:    cfg,
+		client: client,
+	}, nil
+}
+
+// AssignReviewers assigns a list of users and teams as reviewers of a target
+// Pull Request. The implementation assigns one principal per request because
+// mixing existing reviewers in pending review state with new reviewers will
+// result in a no-op without errors thrown.
```
drevell, line 49: Nice comment, thanks for this explanation

drevell (approved):

gjonathanhong:
Synced offline with Mike, we decided we should phase out Guardian's `github` package and instead directly use `go-github`'s client. (Sorry @drevell)

pkg/platform/github.go
```diff
@@ -20,13 +20,25 @@ import (
...
 	gh "github.com/abcxyz/guardian/pkg/github"
 	"github.com/abcxyz/pkg/githubauth"
+	"github.com/abcxyz/pkg/logging"
 )
 
-var _ Platform = (*GitHub)(nil)
+var (
+	_ Platform = (*GitHub)(nil)
+
+	ignoredStatusCodes = map[int]struct{}{
```
drevell, line 34: Did this list come from anywhere? If so, could you link the docs/source/etc? I'm imagining a future maintainer wondering whether to update this or if it needs these exact values.

pkg/platform/github.go
```diff
@@ -76,3 +88,71 @@ func NewGitHub(ctx context.Context, cfg *gh.Config) (*GitHub, error) {
 
 	return g, nil
 }
+
+// RequestReviewers abstracts GitHub's RequestReviewers API with retries.
+func (g *GitHub) requestReviewers(ctx context.Context, reviewer *github.ReviewersRequest) error {
```
drevell, line 93: The mix of singular and plural is a little confusing, should `reviewer` be `reviewers`?

drevell (approved):

pkg/platform/github.go
gjonathanhong: Renamed to `reviewers`. Did the same with `inputs` to `input`, matching the singular `AssignReviewersInput`.

pkg/platform/github.go
gjonathanhong: I added a doc comment. The status codes are defined in each endpoint https://docs.github.com/en/rest/pulls/review-requests?apiVersion=2022-11-28#request-reviewers-for-a-pull-request--status-codes.

I reduced the list to only those that are used by AssignReviewers, but it should be updated when more implementations are added.

gjonathanhong:
Ah, I realized I addressed updates on the wrong branch ☠️

PR #386, opened 2024-08-22
fix: implement reporter

verbanicm:
- Implemented the reporter for the commands that need it
- Refactored the naming of some Reporter funcs
- Added `EntrypointsSummary` func to reporter to handle the output of the `entrypoints` command
- Opted for a `NoopReporter` instead of `LocalReporter` which makes more sense because reporting is about integrating with the "platform" guardian is running on for real-time user feedback. Locally this is apparent in the stdout per command.

To follow: 
- Re-add integration testing
- Default the reporter flag based on platform type


verbanicm:
#### 🔱 Guardian 🔱

Guardian will run for the following directories
**Modified Dirs**
```terraform/project1
terraform/project2```

**Comment by verbanicm** [comment]
#### 🔱 Guardian 🔱

Guardian will run for the following directories
**Modified Dirs**
```
terraform/project1
terraform/project2
```

File: pkg/reporter/github_reporter.go [hunk 1]
```diff
@@ -157,8 +157,29 @@ func (g *GitHubReporter) CreateStatus(ctx context.Context, st Status, p *Params)
...
+	_, err = g.gitHubClient.CreateIssueComment(
+		ctx,
+		g.inputs.GitHubOwner,
+		g.inputs.GitHubRepo,
+		g.inputs.GitHubPullRequestNumber,
+		msg.String(),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to report: %w", err)
+	}
```
sethvargo, line 176:
```suggestion
	if _, err := g.gitHubClient.CreateIssueComment(
		ctx,
		g.inputs.GitHubOwner,
		g.inputs.GitHubRepo,
		g.inputs.GitHubPullRequestNumber,
		msg.String(),
	); err != nil {
		return fmt.Errorf("failed to report: %w", err)
	}
```

pkg/github/config.go
```diff
@@ -43,28 +45,47 @@ type Config struct {
...
+			d.PullRequestBody = event.GetPullRequest().GetBody()
+		}
+	}
 
 	f := set.NewSection("GITHUB OPTIONS")
 
 	f.StringVar(&cli.StringVar{
 		Name:   "github-token",
-		EnvVar: "GITHUB_TOKEN",
+		EnvVar: "GITHUB_REPO_TOKEN",
```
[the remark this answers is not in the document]
verbanicm: So Terraform uses GITHUB_TOKEN for authenticating, there could be a conflict of permissions. If we want to allow for least privilege, we should accept our own token just for writing comments, etc. Open to suggestions, but this seemed the cleanest.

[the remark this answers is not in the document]
verbanicm: Also, GITHUB_TOKEN isnt injected by default in actions. Its stored in the context `github.token` and secrets `secrets.GITHUB_TOKEN`. So, regardless the user will need to pass it to us.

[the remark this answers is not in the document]
verbanicm: Created a guardian-github-token that will be used for guardian, if it is empty it will default to  github_token.

pkg/github/config.go
```diff
@@ -43,28 +46,55 @@ type Config struct {
...
+
 func (c *Config) RegisterFlags(set *cli.FlagSet) {
+	d := &configDefaults{}
+
 	githubContext, _ := githubactions.New().Context()
-	owner, repo := githubContext.Repo()
-	prNumber, _ := githubContext.Event["number"].(int)
+	d.Owner, d.Repo = githubContext.Repo()
+
+	data, _ := json.Marshal(githubContext.Event) //nolint:errchkjson //Shouldnt affect defaults
```
yolocs, line 66:
```suggestion
	data, _ := json.Marshal(githubContext.Event) //nolint:errchkjson //Shouldn't affect defaults
```
Curious what's the purpose of this line. Maybe add a concise note about it.

pkg/github/config.go
```diff
@@ -43,28 +46,55 @@ type Config struct {
...
+		}
+	}
 
 	f := set.NewSection("GITHUB OPTIONS")
 
+	f.StringVar(&cli.StringVar{
+		Name:   "guardian-github-token",
+		EnvVar: "GUARDIAN_GITHUB_TOKEN",
+		Target: &c.GuardianGitHubToken,
+		Usage:  "The GitHub access token for Guardian to make GitHub API calls.",
```
yolocs, line 82: Can't effectively tell the difference between this one and the one below. I suppose they are used for different GH APIs?

yolocs (approved):

pkg/github/config.go
verbanicm: See this comment: https://github.com/abcxyz/guardian/pull/386#discussion_r1729112429

pkg/github/config.go
verbanicm: fixed

pkg/github/config.go
verbanicm: gave better description

PR #390, opened 2024-08-23
Add workflow to test policy changes

gjonathanhong:


.github/workflows/integration.yml
```diff
@@ -177,6 +177,111 @@ jobs:
...
+          version: 'latest'
+
+      - name: 'Evaluate Policy'
+        id: 'opa_eval'
+        env:
+          DIRECTORY: '${{ matrix.working_directory }}'
+          TEMP_DIR: '${{ runner.temp }}'
+        shell: 'bash'
+        run: |-
+          terraform -chdir=${DIRECTORY} show -json tfplan.binary > tfplan.json
```
verbanicm, line 267: I wonder if we should build this into the plan command to just write the json file too.

.github/workflows/integration.yml
```diff
@@ -177,6 +177,111 @@ jobs:
         run: |-
           guardian plan -dir="${DIRECTORY}" -storage="gcs://${GUARDIAN_BUCKET_NAME}"
 
+      # - name: 'Aggregate PR Approver Data'
```
verbanicm, line 180: Im sure youre thinking of this but, this could be part of the platform impl and generated during policy command.

pkg/commands/policy/policy.go
```diff
@@ -127,18 +127,19 @@ func (c *PolicyCommand) Process(ctx context.Context) error {
 
 			merr = errors.Join(merr, fmt.Errorf("failed: \"%s\" - %s", k, m.Message))
 		}
+	}
 
+	if len(teams) > 0 || len(users) > 0 {
```
verbanicm, line 132: See comment on other PR

PR #392, opened 2024-08-23
Add check for non-zero missing approvals

gjonathanhong:


pkg/commands/policy/policy.go
```diff
@@ -127,18 +127,19 @@ func (c *PolicyCommand) Process(ctx context.Context) error {
...
-		Users: users,
-	}); err != nil {
-		return fmt.Errorf("failed to assign reviewers: %w", err)
+		if _, err := c.platform.AssignReviewers(ctx, &platform.AssignReviewersInput{
+			Teams: teams,
+			Users: users,
+		}); err != nil {
+			return fmt.Errorf("failed to assign reviewers: %w", err)
+		}
 	}
```
verbanicm, line 143: I think this would be more idiomatic...

- Short circuit if you have non-nil merr from the policy eval
- Otherwise if no length return nil

```suggestion
  if merr != nil {
		return merr //nolint:wrapcheck // Want passthrough
  }

	if len(teams) == 0 && len(users) == 0 {
	  return nil
	}
	
	logger.DebugContext(ctx, "found missing approvals",
		"teams", teams,
		"users", users,
	)
	if _, err := c.platform.AssignReviewers(ctx, &platform.AssignReviewersInput{
		Teams: teams,
		Users: users,
	}); err != nil {
		return fmt.Errorf("failed to assign reviewers: %w", err)
	}
	
	return nil
```

gjonathanhong: > Short circuit if you have non-nil merr from the policy eval

I would prefer to keep merr at the end so that reviewers are assigned before returning the policy result

verbanicm: ah makes sense

verbanicm (approved):

pkg/commands/policy/policy.go
```diff
@@ -127,13 +127,16 @@ func (c *PolicyCommand) Process(ctx context.Context) error {
...
 			merr = errors.Join(merr, fmt.Errorf("failed: \"%s\" - %s", k, m.Message))
 		}
+	}
 
-		logger.DebugContext(ctx, "found missing approvals",
-			"teams", teams,
-			"users", users,
-		)
+	if len(teams) == 0 && len(users) == 0 {
+		return nil
```
drevell, line 133: Could you add a code comment here explaining why merr is ignored in this case?

drevell (changes requested):

pkg/commands/policy/policy.go
gjonathanhong: Thanks, we actually want to return `merr` and left a comment explaining why

drevell (approved):

PR #394, opened 2024-08-26
Save plan file as JSON file in plan command

gjonathanhong:
This change writes the plan to a JSON file, in addition to the existing plan binary, so that it can be parsed by other commands (specifically the `policy` command). 

Added a new `output-dir` flag to specify the directory that plan binary and json files are written to. This is helpful for writing to temp directories in unit tests, so that any written files are cleaned up after running unit tests.

### DIFF
--- a/pkg/commands/plan/plan.go
+++ b/pkg/commands/plan/plan.go
@@ -326,7 +326,7 @@ func (c *PlanCommand) terraformPlan(ctx context.Context) (*RunResult, error) {
 	stdout.Reset()
 	stderr.Reset()
 
-	util.Headerf(c.Stdout(), "Saving output to JSON file")
+	util.Headerf(c.Stdout(), "Writing Plan to Local JSON File")
 	_, err = c.terraformClient.Show(ctx, &stdout, multiStderr, &terraform.ShowOptions{
 		File:    pointer.To(c.planFilename),
 		NoColor: pointer.To(true),
--- a/pkg/commands/plan/plan.go
+++ b/pkg/commands/plan/plan.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -64,15 +65,17 @@ type RunResult struct {
 type PlanCommand struct {
 	cli.BaseCommand
 
-	directory     string
-	childPath     string
-	planFilename  string
-	storagePrefix string
+	directory        string
+	childPath        string
+	planFilename     string
+	planJSONFilename string
+	storagePrefix    string
 
 	platformConfig platform.Config
 
 	flags.CommonFlags
 
+	flagOutputDir            string
 	flagDestroy              bool
 	flagStorage              string
 	flagAllowLockfileChanges bool
@@ -135,6 +138,11 @@ func (c *PlanCommand) Flags() *cli.FlagSet {
 		Usage:   "The duration Terraform should wait to obtain a lock when running commands that modify state.",
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:    "output-dir",
+		Target:  &c.flagOutputDir,
+		Example: ".",
+	})
 	return set
 }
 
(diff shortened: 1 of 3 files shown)


pkg/commands/plan/plan.go
```diff
@@ -322,6 +326,40 @@ func (c *PlanCommand) terraformPlan(ctx context.Context) (*RunResult, error) {
...
+
+	var data any
+	if err := json.Unmarshal([]byte(stdout.String()), &data); err != nil {
+		return &RunResult{hasChanges: hasChanges}, fmt.Errorf("failed to unmarshal json plan: %w", err)
+	}
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(data); err != nil {
+		return &RunResult{hasChanges: hasChanges}, fmt.Errorf("failed to encode json: %w", err)
+	}
```
verbanicm, line 353: Is there a reason for the unmarshal and encoding, seems unnecessary? Can't we just write the string bytes to the file?

```go
if err := os.WriteFile("test", []byte(stdout.String()), 0o600); err != nil {
  // handle err
}
```

pkg/commands/plan/plan.go
```diff
@@ -322,6 +326,40 @@ func (c *PlanCommand) terraformPlan(ctx context.Context) (*RunResult, error) {
 	stdout.Reset()
 	stderr.Reset()
 
+	util.Headerf(c.Stdout(), "Saving output to JSON file")
```
verbanicm, line 329: nit: might be confusing between GCS and local for "saving"...

```suggestion
	util.Headerf(c.Stdout(), "Writing Local Plan to JSON File")
```

pkg/commands/plan/plan.go
gjonathanhong: It's to validate that `stdout.String()` is valid JSON, in case we forget to clear it prior to running the `terraform show` command.

verbanicm: Seems wasteful, should we just create a new `var jsonOut strings.Builder` to prevent that issue?

pkg/commands/plan/plan.go
```diff
@@ -322,6 +325,30 @@ func (c *PlanCommand) terraformPlan(ctx context.Context) (*RunResult, error) {
 	stdout.Reset()
 	stderr.Reset()
 
+	if !c.skipJSONPlanFile {
```
verbanicm, line 328: Will this be behind a flag? What is this for since the user cant change this?

pkg/commands/plan/plan.go
```diff
@@ -322,6 +325,30 @@ func (c *PlanCommand) terraformPlan(ctx context.Context) (*RunResult, error) {
...
+			JSON:    pointer.To(true),
+		}); err != nil {
+			return &RunResult{
+				commentDetails: stderr.String(),
+				hasChanges:     hasChanges,
+			}, fmt.Errorf("failed to terraform show: %w", err)
+		}
+
+		planBasename := strings.TrimSuffix(c.planFilename, path.Ext(c.planFilename))
+		jsonFilepath := path.Join(c.childPath, planBasename+".json")
```
verbanicm, line 343: We should figure out a better way for this, this feels hacky, but we can address in a follow up.

verbanicm (approved):

pkg/commands/plan/plan.go
gjonathanhong: it's moreso for skipping the json file in unit tests, because it creates a file for every time `go test` is run. I'm trying to see if there's a better way to override the file location to `t.TempDir()` instead of `c.childDir`

verbanicm: ah yes, that is tough since were using testdata, I was thinking we should write the out files to a temp dir anyways....so might be time to factor that in.

pkg/commands/plan/plan.go
gjonathanhong: I changed this to mimick the `planFilename` pattern for now

pkg/commands/plan/plan.go
gjonathanhong: Added a new flag `output-dir` where the binary and json file will be written to. This can be either a child directory or not, but the storage client will continue to use the existing `directory` path.

pkg/commands/plan/plan.go
```diff
@@ -133,6 +138,11 @@ func (c *PlanCommand) Flags() *cli.FlagSet {
 		Usage:   "The duration Terraform should wait to obtain a lock when running commands that modify state.",
 	})
 
+	f.StringVar(&cli.StringVar{
```
verbanicm, line 141: nit: can you add a usage description?

verbanicm (approved):

PR #396, opened 2024-08-27
fix: cleanup plan command

verbanicm:
Also removed the override env vars from the child process to being passed in by the caller.

/cc @gjonathanhong

### DIFF
--- a/pkg/terraform/run.go
+++ b/pkg/terraform/run.go
@@ -26,7 +26,7 @@ import (
 // overrideEnvVars are the environment variables to inject into the Terraform
 // child process, no matter what the user configured. These take precedence
 // over all other configurables.
-var overrideEnvVars = []string{
+var defaultOverrideEnvVars = []string{
 	"GOOGLE_TERRAFORM_USERAGENT_EXTENSION=" + version.UserAgent,
 	"TF_APPEND_USER_AGENT=" + version.UserAgent,
 }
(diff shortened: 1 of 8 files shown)


pkg/terraform/run.go
```diff
@@ -17,10 +17,20 @@ package terraform
...
 
+	"github.com/abcxyz/guardian/internal/version"
 	"github.com/abcxyz/guardian/pkg/child"
 )
 
+// overrideEnvVars are the environment variables to inject into the Terraform
+// child process, no matter what the user configured. These take precedence
+// over all other configurables.
+var defaultOverrideEnvVars = []string{
+	"GOOGLE_TERRAFORM_USERAGENT_EXTENSION=" + version.UserAgent,
```
drevell, line 30: I know this is out of scope for this PR. Perhaps someday, a cleaner API for environment variables wouldn't rely on parsing the equals sign (as happens in the `environ` function), but would pass keys and values separately.

It could be a map, like:
```
myEnvVars := map[string]string{
  "MY_ENV_VAR_1": "some_value",
```

Or a list of pairs, like:
```
myEnvVars := [][]string{{"ENV_VAR_1", "value1"}, {"ENV_VAR_2", "value2"}
```

When I saw these equals signs I was confused at what could possibly be happening, until I discovered that we implemented our own little parser for environment vars using `strings.SplitN` which seems a bit much.

drevell (approved):

pkg/terraform/run.go
verbanicm: Could be cleaner, just followed existing impl. We can discuss offline.

PR #397, opened 2024-08-27
Add GetLatestApprovers to Platform interface

gjonathanhong:


pkg/platform/github.go
```diff
@@ -150,6 +154,96 @@ func (g *GitHub) AssignReviewers(ctx context.Context, input *AssignReviewersInpu
...
+			Nodes []team
+		} `graphql:"teams(first: 100)"`
+	} `graphql:"organization(login: $owner)"`
+}
+
+// GetLatestApprovers retrieves the users whose latest review for a pull request
+// is an approval. It also returns the teams and subteams that the user
+// approvers are members of to indicate approval on behalf of those teams. Note,
+// if a reviewer leaves a comment OR requests changes after a previous approval,
+// then the reviewer will not be included in the "latest" approvers list.
```
verbanicm, line 195: Is this the behavior we want? I know we've touched on this, but I would think any approval is an approval.

pkg/platform/github.go
```diff
@@ -41,8 +42,9 @@ var (
 
 // GitHub implements the Platform interface.
 type GitHub struct {
-	cfg    *gh.Config
-	client *github.Client
+	cfg           *gh.Config
+	client        *github.Client
+	graphQLClient *githubv4.Client
```
verbanicm, line 47: nit: I think the "appropriate" naming would be graphqlClient

pkg/platform/github.go
gjonathanhong: I just tested the behavior. It looks like a comment will not change the status of the review, so a prior approval will still be in effect. If the approval is dismissed or followed by a requested changes, then the approval won't be counted.

That does seem like the behavior we want.

verbanicm: +1, thanks!

pkg/platform/github.go
```diff
@@ -150,6 +154,97 @@ func (g *GitHub) AssignReviewers(ctx context.Context, input *AssignReviewersInpu
...
+	hasApproved := make(map[string]struct{}, len(approversQuery.Repository.PullRequest.LatestReviews.Nodes))
+	for _, review := range approversQuery.Repository.PullRequest.LatestReviews.Nodes {
+		if review.State == "APPROVED" {
+			result.Users = append(result.Users, review.Author.Login)
+			hasApproved[review.Author.Login] = struct{}{}
+		}
+	}
+
+	var teamQuery organizationTeamsAndMembershipsQuery
+	if err := g.graphqlClient.Query(ctx, &teamQuery, map[string]any{
```
verbanicm, line 225: followup: should we paginate this? I would guess some orgs might have over 100 teams, users, etc

Example: https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/review/commit_review_status.go

verbanicm (approved):

PR #401, opened 2024-08-28
Add policy fetch-data command

gjonathanhong:
This command fetches data used as context for policy evaluation and writes it to a local file named `guardian_policy_context.json`. It currently only aggregates the pull request approvers data.

### DIFF
--- a/pkg/commands/policy/fetch_data.go
+++ b/pkg/commands/policy/fetch_data.go
@@ -18,12 +18,20 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"path"
 
 	"github.com/abcxyz/guardian/pkg/platform"
+	"github.com/abcxyz/guardian/pkg/util"
 	"github.com/abcxyz/pkg/cli"
 )
 
 var _ cli.Command = (*FetchDataCommand)(nil)
+
+const (
+	ownerReadWritePerms = 0o600
+	policyDataFilename  = "guardian_policy_context.json"
+)
 
 // FetchDataCommand implements cli.Command. It fetches and aggregates data for
 // a target platform to be used for policy evaluation.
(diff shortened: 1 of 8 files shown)


pkg/commands/policy/fetch_data.go
```diff
@@ -0,0 +1,85 @@
...
+		return fmt.Errorf("failed to create platform: %w", err)
+	}
+	c.platform = platform
+
+	return c.Process(ctx)
+}
+
+// Process handles the main logic for aggregating data for the target platform.
+func (c *FetchDataCommand) Process(ctx context.Context) error {
+	approvers, err := c.platform.GetLatestApprovers(ctx)
```
verbanicm, line 76: I think in my head i saw this as a `platform.GetPolicyData` and it can call all the APIs it needs, aggregate github configs, etc to generate a well known payload.

Ideally this would be something like (this is totally off the cuff, but I would put some thought into it, its fine to start with just the approvals for now):

```json
{
  "github": {
    "actor": {
      "id": 123123123,
      "login": "username",
      "repo_permissions": {
        "permission": "admin",
        "role_name": "admin"
      }
    },
    "approvals": {
      "users": [],
      "teams": []
    },
    "event": {
      // put the json event payload
    },
    "workflow": {
      // add other workflow information
    }
  }
}

```

verbanicm: At least for now...(

```json
{
  "github": {
    "pull_request_approvals": {
      "users": [],
      "teams": []
    },
  }
}

```

drevell (approved):

kevya-google (approved):

pkg/commands/policy/fetch_data_test.go
```diff
@@ -0,0 +1,122 @@
...
+				},
+			}
+			outFilepath := path.Join(outDir, policyDataFilename)
+
+			err := c.Process(ctx)
+			if diff := testutil.DiffErrString(err, tc.wantErr); diff != "" {
+				t.Errorf("unexpected result; (-got,+want): %s", diff)
+			}
+			if err != nil {
+				// Skip rest of checks because the file won't exist.
```
sailorlqh, line 103: Do we want to make sure the file is not created here?

PR #403, opened 2024-08-30
Add pull_request_target event default values

gjonathanhong:


verbanicm (approved):
Im pretty sure pull_request and pull_request_target are just pull request events....we should see if we can consolidate this behind a single if condition.

gjonathanhong:
> Im pretty sure pull_request and pull_request_target are just pull request events....we should see if we can consolidate this behind a single if condition.

I was testing in a separate PR and the defaults weren't being set, hoping this will fix it

PR #405, opened 2024-09-10
fix: destroy from modifiers

verbanicm:
Guardian shouldn't care if a directory was removed for destroy, there could be a case where we want to just destroy the resources and then run an apply for a clean slate.

Rather the modifier content will be used to determine a destroy phase using the base branch from a pull request and `HEAD~1` from `main` for apply on push.

### DIFF
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -22,10 +22,8 @@ import (
 	"flag"
 	"fmt"
 	"io/fs"
-	"os"
 	"path"
 	"slices"
-	"strings"
 
 	"golang.org/x/exp/maps"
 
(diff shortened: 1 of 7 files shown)


gjonathanhong (approved):

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -198,44 +196,45 @@ func (c *EntrypointsCommand) Process(ctx context.Context) error {
 	}
 
 	var modifiedEntrypoints []string
-	var removedEntrypoints []string
 
 	for _, dir := range c.flagDir {
 		dirAbs, err := util.PathEvalAbs(dir)
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				// ignore mising directories, we will just return nothing
```
yolocs, line 204:
```suggestion
				// ignore missing directories, we will just return nothing
```

pkg/platform/local.go
```diff
@@ -43,7 +64,7 @@ func (l *Local) GetPolicyData(ctx context.Context) (*GetPolicyDataResult, error)
 
 // ModifierContent is a no-op.
```
yolocs, line 65: Update?

yolocs (approved):

PR #411, opened 2024-10-10
Metrics

pdewilde:


pdewilde:
Thoughts on testing metrics?

In my previous team, we would have absolutely tested whether metrics are being incremented. The current integration tests aren't really set up for that sort of testing, and adding to unit tests would mostly be a change detector test, not certain if that is what we want.

cmd/guardian/main.go
```diff
@@ -135,6 +141,33 @@ func main() {
...
+			defer handler()
+			panic(r)
+		}
+	}()
+
+	cleanup := metricswrap.WriteMetric(ctx, mClient, "runs", 1)
+	defer cleanup()
+
+	// This will cause a synchronous metrics call.
+	defer func() {
```
sethvargo, line 164: This defer will run before the cleanup defer; is that intended?

cmd/guardian/main.go
```diff
@@ -135,6 +141,33 @@ func main() {
 
 func realMain(ctx context.Context) error {
 	setLogEnvVars()
+
```
sethvargo, line 144: Should we move this into a helper function (like we did with `setLogEnvVars`) just to keep main clean?

internal/metricswrap/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+// so users will see at most 500ms runtime from this in most cases.
+const defaultMetricsTimeout = 500 * time.Millisecond
+
+// WriteMetric is an async wrapper for metrics.WriteMetric.
+// It returns a function that blocks on completion and handles any errors.
+func WriteMetric(ctx context.Context, client *metrics.Client, name string, count int64) func() {
+	ctx, done := context.WithTimeout(ctx, defaultMetricsTimeout)
+	errCh := make(chan error, 1)
+	go func() {
+		defer done()
```
sethvargo, line 35: I think you want to move this up to L33 (right after the context is created)

internal/metricswrap/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+		defer close(errCh)
+		errCh <- client.WriteMetric(ctx, name, count)
+	}()
+
+	return func() {
+		err := <-errCh
+		if err != nil {
+			logger := logging.FromContext(ctx)
+			logger.DebugContext(ctx, "Metric writing failed.", "err", err)
```
sethvargo, line 44:
```suggestion
			logger.DebugContext(ctx, "failed to write metric", "err", err)
```

sethvargo, line 44: Nit, you can squish this since it's only used once:

```go
logging.FromContext(ctx).DebugContext(ctx, "failed to write metric", "err", err)
```

pkg/commands/drift/command.go
```diff
@@ -117,6 +119,10 @@ func (c *DetectIamDriftCommand) Flags() *cli.FlagSet {
 }
 
 func (c *DetectIamDriftCommand) Run(ctx context.Context, args []string) error {
+	mClient := metrics.FromContext(ctx)
+	cleanup := metricswrap.WriteMetric(ctx, mClient, "command_iam_detect_drift", 1)
+	defer cleanup()
```
sethvargo, line 124:
```suggestion
	defer metricswrap.WriteMetric(ctx, mClient, "command_iam_detect_drift", 1)()
```

Is there a reason that won't work?

sethvargo (commented):
I'm not sure I like the `metricswrap` @pdewilde. Can you share more about why you think we need it?

README.md
```diff
@@ -306,3 +306,24 @@ in the [`abc.templates`](abc.templates) folder.
...
 [template installation instructions](abc.templates/README.md#install-drift-detection-workflows)
 in the [`abc.templates`](abc.templates) folder.
+
+## Metrics
+Google collects usage statics. No identifiable data is collected.
+Metrics are collected using [abcxyz/abc-updater](https://github.com/abcxyz/abc-updater).
+
+Currently, data is collected on:
+- Count of total invocations
+- Count of each sub-command (render, describe, upgrade, ect)
```
verbanicm, line 316: These are abc commands

README.md
```diff
@@ -306,3 +306,24 @@ in the [`abc.templates`](abc.templates) folder.
 To use Guardian drift detection in your repository, see the
 [template installation instructions](abc.templates/README.md#install-drift-detection-workflows)
 in the [`abc.templates`](abc.templates) folder.
+
+## Metrics
+Google collects usage statics. No identifiable data is collected.
+Metrics are collected using [abcxyz/abc-updater](https://github.com/abcxyz/abc-updater).
```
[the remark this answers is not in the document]
pdewilde: Will also make change on abcxyz/abc.

cmd/guardian/main.go
```diff
@@ -41,6 +44,9 @@ const (
 	defaultLogLevel  = "warn"
 	defaultLogFormat = "json"
 	defaultLogDebug  = "false"
+	// Shorter since nothing can be done in parallel due to it starting after
```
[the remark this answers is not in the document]
pdewilde: // Shorter than default metrics timeout since nothing can be done in parallel
	// due to it starting after program logic finishes.

cmd/guardian/main.go
pdewilde: I guess that is a good point. I was thinking the order didn't really matter but we probably want the runtime to be all inclusive.

cmd/guardian/main.go
pdewilde: Probably, the main issue is handling the defer statements in the correct order, we would need to return a function for the caller to defer. I'll think about if metricswrap is really needed.

cmd/guardian/main.go
```diff
@@ -135,6 +141,33 @@ func main() {
 
 func realMain(ctx context.Context) error {
 	setLogEnvVars()
+
+	start := time.Now()
+	mClient, err := metrics.New(ctx, version.Name, version.Version)
+	if err != nil {
+		fmt.Printf("metric client creation failed: %v\n", err)
```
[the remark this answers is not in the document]
pdewilde: Good point. I think its not usable or actionable to an end user. I'll pop the logging context before client creation so this can be a debug log.

README.md
pdewilde: Good catch!

cmd/guardian/main.go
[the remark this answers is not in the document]
pdewilde: Edited to clarify a bit.

In a world where `metricswrap` exists, all metrics calls can be done concurrently with the body of the function they are in, while runtime by definition must run at the end of the program, so latency will be felt sequentially by users.

cmd/guardian/main.go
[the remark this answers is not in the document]
pdewilde: Changed, also made change in `abc`

internal/metricswrap/metrics.go
pdewilde: Wouldn't that cancel the context when `WriteMetric` returns? Maybe it would make sense to create the error channel before the context is created, but done() needs to happen within the go func

internal/metricswrap/metrics.go
pdewilde: Changed, same in `abc`

pkg/commands/drift/command.go
pdewilde: That would work fine. I guess I thought it was easier to reason about what is happening by aliasing the returned function rather than calling it directly.

pdewilde:
@sethvargo 
> I'm not sure I like the `metricswrap` @pdewilde. Can you share more about why you think we need it?

Yeah, this is something we discussed in the past including directly in abc-updater but decided to let consumers of the metrics library write concurrency themselves.

I was copying my metrics code from `abc` where I think it makes more sense, as it is used synchronously by users where latency really matters to UX. It is also true that a user of `abc` isn't necessarily close to `us-central1` (the only location the metrics server is currently deployed) and may see significant latency for each round-trip (I recall some internal tools at by old job being borderline unusable when I was in Croatia).

Here, I guess there is an argument to make that the spin-up of the runner and calls to refresh TF state are likely much longer than calls to the metrics server, and I suspect most GH Actions be running form the continental US where calls to the metrics server will have lower latency. The only counter argument here is that pipeline speed is very important for developer velocity for Guardian, as users cannot access state to test things locally.

If we don't expect to be adding many more metrics in the future, I'm fine to take metricswrap out of Guardian. If @verbanicm or another developer forsees wanting to add more than the 3 we currently have, its probably worth discussing either adding batching to the API or keeping metricswrap.

sethvargo:
These are all reasonable, but I'd strongly prefer to keep things simple until there's a recognized problem. Since guardian is almost always running in server-side infrastructure, the latency concern seems largely mtigated

pdewilde:
I removed the async stuff but left in a wrapper for some quality of life improvements. (It fetches client from context, creates context with timeout, sends errors to debug, and calls the done function on the context)

Might be worth adding some into the metrics library (set default time in particular is nice).

cmd/guardian/main.go
pdewilde: Was able to factor some of the setup out.

README.md
```diff
@@ -306,3 +306,26 @@ in the [`abc.templates`](abc.templates) folder.
...
+Along with each metric, the following metadata is recorded:
+- Application version
+- Installation time with minute granularity
+
+Metrics data is retained for 24 months.
+
+You can **opt-out** of collection by setting the following environment variable:
+```shell
+export GUARDIAN_NO_METRICS=TRUE
+```
```
sethvargo, line 331: I think you can cut this since it's in the first paragraph now.

internal/metricswrap/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+// WriteMetric is a sync wrapper for metrics.WriteMetric.
+// It handles client retrival, timeouts, and errors.
+func WriteMetric(ctx context.Context, name string, count int64) {
+	client := metrics.FromContext(ctx)
+	ctx, done := context.WithTimeout(ctx, defaultMetricsTimeout)
+	defer done()
+	if err := client.WriteMetric(ctx, name, count); err != nil {
+		logging.FromContext(ctx).DebugContext(ctx, "failed to write metric", "err", err)
+	}
+}
```
sethvargo, line 36: So this helper:

1. Adds a timeout for writing
2. Logs on error writing

pkg/commands/workflows/remove_comments.go
```diff
@@ -55,6 +56,8 @@ func (c *RemoveGuardianCommentsCommand) Flags() *cli.FlagSet {
 }
 
 func (c *RemoveGuardianCommentsCommand) Run(ctx context.Context, args []string) error {
+	metricswrap.WriteMetric(ctx, "command_workflows_remove_guardian_comments", 1)
```
sethvargo, line 59: I like this developer API better. As a follow-up, I wonder if we could automatically export these metrics magically in the CLI package.

sethvargo (approved):

internal/metricswrap/metrics.go
pdewilde: Correct. It also extracts the client from the context

pkg/commands/workflows/remove_comments.go
pdewilde: That's an interesting thought. Current design of metrics does still require allow listing each metric so you would still have an extra step.

As for how to expose in CLI package, I can see a few options.
1. Have command factory inject the command(s) that caused this command to run into the cli.Command interface. []string. For example run would just be ["run"] while iam cleanup would be ["iam", "cleanup"]. 

The main drawback to which approach is if you have multiple possible paths to the same Command (for example `abc` dropped the `templates` in `abc templates render`, but had to keep it as an option for backwards compatibility.

2. Have cli.Command expose its type somehow. Either with an explicit field or getting struct name via reflection.

Refelction feels like a bit of a smell, so probably not the best choice.



As far as dispatching the metrics go, if you don't want to be explicit, I would think you would probably want something more generic, like adding an arbitrary function to the context which gets called before Run().

sethvargo: I was thinking more generic actually. The CLI can introduce things like "pre" and "post" actions (e.g. `BeforeRun`, `AfterRun`), and the main registration could hook into that. We already do this a bit with `AfterParse` FWIW, but you could register metrics as a hook.

verbanicm (approved):

sethvargo (approved):

PR #412, opened 2024-10-25
refactor: generalize status message helpers for Reporter types

gjonathanhong:
This change moves helper methods and common types out from `github_reporter.go` to `reporter.go`. They were previously tied to the GitHubReporter type, but made more sense as common types/methods that shouldn't change between platforms. This will allow us to generate status messages that are consistent across different code review platforms, instead of reimplementing them each time.


drevell (approved):
Nice PR description 👍

PR #413, opened 2024-10-25
Add FileReporter

gjonathanhong:
This change provides an option to report the status or entrypoints summary comment to a local file on disk.

### DIFF
--- a/pkg/reporter/file_reporter.go
+++ b/pkg/reporter/file_reporter.go
@@ -1,3 +1,17 @@
+// Copyright 2024 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pkg/reporter/file_reporter.go
```diff
@@ -0,0 +1,80 @@
...
+func (f *FileReporter) EntrypointsSummary(ctx context.Context, params *EntrypointsSummaryParams) error {
+	logger := logging.FromContext(ctx)
+
+	msg, err := entrypointsSummaryMessage(params, "")
+	if err != nil {
+		return fmt.Errorf("failed to generate entrypoints summary message: %w", err)
+	}
+
+	if err := os.WriteFile(entrypointsSummaryFilename, []byte(msg.String()), ownerReadWritePerms); err != nil {
+		return fmt.Errorf("failed to write file")
```
sailorlqh, line 68: Should you include and warp the error info here?
Or just log it?

sailorlqh (commented):
I assume the test will be included in the follow up PR?

sailorlqh (approved):

pkg/reporter/file_reporter.go
gjonathanhong: Thanks for catching

gjonathanhong:
> I assume the test will be included in the follow up PR?

Yes, will do in follow up :)

PR #417, opened 2024-11-11
Add UserAccessLevel in GetPolicyData result

gjonathanhong:
This change adds `GetUserRepoPermissions` to the Platform interface, which adds the user's access level for a repository as additional context in policy evaluations. This also changes `platform.GetPolicyData()` to skip fetching Pull Request Approvers if a pull request number is not provided.

```javascript
{
  github: {
    pull_request_approvers: {
      // ...
    },
    user_access_level: "admin",
  }
}
```

### DIFF
--- a/pkg/github/config.go
+++ b/pkg/github/config.go
@@ -77,6 +77,7 @@ func (c *Config) RegisterFlags(set *cli.FlagSet) {
 		if err := json.Unmarshal(data, &event); err == nil {
 			d.PullRequestNumber = event.GetNumber()
 			d.PullRequestBody = event.GetPullRequest().GetBody()
+			d.WorkflowUser = event.GetSender().GetLogin()
 		}
 	}
 	if githubContext.EventName == "pull_request_target" {
@@ -84,6 +85,7 @@ func (c *Config) RegisterFlags(set *cli.FlagSet) {
 		if err := json.Unmarshal(data, &event); err == nil {
 			d.PullRequestNumber = event.GetNumber()
 			d.PullRequestBody = event.GetPullRequest().GetBody()
+			d.WorkflowUser = event.GetSender().GetLogin()
 		}
 	}
 	if githubContext.EventName == "workflow_dispatch" {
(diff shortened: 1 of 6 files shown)


pkg/github/config.go
```diff
@@ -221,4 +231,13 @@ GITHUB_TOKEN.`,
 		Usage:  "The GitHub SHA.",
 		Hidden: true,
 	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "github-workflow-user",
```
verbanicm, line 236: Should we just use `GITHUB_ACTOR` instead?

pkg/github/config.go
```diff
@@ -75,13 +77,21 @@ func (c *Config) RegisterFlags(set *cli.FlagSet) {
...
 		if err := json.Unmarshal(data, &event); err == nil {
 			d.PullRequestNumber = event.GetNumber()
 			d.PullRequestBody = event.GetPullRequest().GetBody()
+			d.Actor = event.GetSender().GetLogin()
+		}
+	}
+	if githubContext.EventName == "workflow_dispatch" {
+		var event github.WorkflowDispatchEvent
+		if err := json.Unmarshal(data, &event); err == nil {
+			d.Actor = event.GetSender().GetLogin()
```
verbanicm, line 94: I think this is automatically injected into the environment, is there a difference or does it differ in certain scenarios where we would want sender vs actor?

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables:~:text=GITHUB_ACTOR,example%2C%20octocat.

verbanicm (approved):

pkg/github/config.go
gjonathanhong: It looks like GitHub sender is guaranteed to be a user, but actor can be both a user or an app. Not sure if it matters, so using actor for now

verbanicm: sgtm

PR #419, opened 2024-11-13
fix: entrypoints command returns empty json array instead of null

gjonathanhong:


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -306,6 +306,9 @@ func (c *EntrypointsCommand) writeOutput(cwd string, results []string) error {
 		results[k] = childPath
 	}
 
+	if results == nil {
```
sailorlqh, line 309: What do you think if we do this nil check at line 299, and `return nil` if `results == nil`?

sailorlqh (approved):

pkg/commands/entrypoints/entrypoints.go
verbanicm: We still need to write it as json below, so we cant exit early.

What if we just initialize the initial slice to 0 length on line 192.

```go
	modifiedEntrypoints := make([]string, 0)
```

gjonathanhong: I had considered that, but was conflicted because the Go style guide prefers initializing with a nil slice over empty slice. I suppose this could be an exception since we're encoding to JSON.

sailorlqh (approved):

sethvargo:
This is fine, but it's still error-prone. It would be better if we used a custom encoder. This has coupled the business logic with the presentation logic. https://blog.gopheracademy.com/advent-2016/advanced-encoding-decoding/

PR #421, opened 2024-11-13
Add custom json encoder for guardian entrypoints

gjonathanhong:


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -177,6 +177,18 @@ func (c *EntrypointsCommand) Run(ctx context.Context, args []string) error {
...
+// Entrypoints is a custom type that implements the json.Marshaler interface.
+type Entrypoints []string
+
+// MarshalJSON implements json.Marshaler.MarshalJSON.
+func (e Entrypoints) MarshalJSON() ([]byte, error) {
+	b, err := json.Marshal(NewResult(&e))
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal entrypoints: %w", err)
+	}
+	return b, nil
```
sethvargo, line 189: You can also handle the empty case here:

```go
if len(e) == 0 {
  return []byte("[]"), nil
}
return json.Marshal(any(e)) // I think the cast to `any` works to prevent the recursive loop, untested
```

But my original point was that we should move this into the "presentation" layer (MVC). So I would expect the `reporterClient` to know how to handle this presentation. If we're writing JSON vs YAML vs an API call, the representation might be different. That's why I think it needs to be pushed down a layer. Does that make sense?

PR #425, opened 2024-11-15
refactor: Policy Fetch-Data command returns only teams that the user is a member of

gjonathanhong:
This change improves the structure of the data available for authoring new policies. 

Now, the user's teams are included in the payload of the `github.actor`, instead of previously all teams and members of the org.
```
{
  "github": {
    "actor": {
      "username": "abcdefg",
      "teams": ["team1", "team2"]
    }
  }
}
```

To make this possible, instead of a single big query, `GetTeamMemberships` was refactored into `GetUserTeamMemberships`, invoked once per user.

### DIFF
--- a/pkg/commands/policy/fetch_data_test.go
+++ b/pkg/commands/policy/fetch_data_test.go
@@ -157,7 +157,6 @@ func TestFetchData_Process(t *testing.T) {
 					UserApprovers:    tc.users,
 					UserAccessLevel:  tc.userAccessLevel,
 					IsPullRequest:    tc.isPullRequest,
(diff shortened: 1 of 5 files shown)


verbanicm:
I am confused by this, we return a list of teams the actor belongs to and all their members? I thought we would just list the teams the user belongs to, what is the use case to know all the members of that team?

gjonathanhong:
Ready for review 🙏 

**Note:** The scope of this PR has changed completely.

sailorlqh (approved):

drevell (approved):

PR #431, opened 2024-11-19
Change format and ordering of policy enforcement comment

gjonathanhong:


pkg/commands/policy/enforce.go
```diff
@@ -211,6 +203,15 @@ func (c *EnforceCommand) EnforceMissingApprovals(ctx context.Context, b *strings
 		"teams", teams,
 		"users", users,
 	)
+
+	fmt.Fprint(b, "- **Missing approvals from one of**:\n")
```
gjonathanhong, line 207: Moved here, because it should report the status after the aggregation of users and teams.

pkg/commands/policy/enforce.go
```diff
@@ -153,7 +153,7 @@ func (c *EnforceCommand) Process(ctx context.Context) error {
 
 		if violation != nil {
 			// Prints policy name followed by the violations found.
-			fmt.Fprintf(&b, "#### %s\n", k)
+			fmt.Fprintf(&b, "#### Policy: `%s`\n", k)
```
gjonathanhong, line 156: For easier identification of the policy name in the status comment

verbanicm (approved):

PR #432, opened 2024-11-19
fix: return only teams actor is a member of

gjonathanhong:
There was a bug that ended up returning all teams in the organization as the teams the github actor is a member of. This properly returns only the teams they are members of, both direct and inherited.


verbanicm (approved):

drevell:
Just based on the PR description, this area seems in need of some unit tests.

PR #441, opened 2024-12-03
Implement GitLabReporter Status

gjonathanhong:


pkg/platform/config/gitlab.go
gjonathanhong, line 1: The location of this file seems a bit odd to me. I'm thinking about making a `gitlab` package under `/pkg` and also create a mock client for tests.

Open to suggestions 🤷

gjonathanhong:
obsolete, went with #443 instead

PR #443, opened 2024-12-04
Add CommentStatus to Platform interface with GitHub implementation

gjonathanhong:
This change is part of a multi-PR change to migrate all Reporter functionality to the Platform interface. 

The motivation is to report based on the platform type, instead of allowing a mix and match between the platform type and reporter type. This will also help avoid cyclical dependencies stemmed from sharing configs.

### DIFF
--- a/pkg/platform/reporter.go
+++ b/pkg/platform/reporter.go
@@ -21,11 +21,11 @@ import (
 )
 
 const (
-	StatusSuccess         Status = Status("SUCCESS")          //nolint:errname // Not an error
-	StatusFailure         Status = Status("FAILURE")          //nolint:errname // Not an error
-	StatusNoOperation     Status = Status("NO CHANGES")       //nolint:errname // Not an error
-	StatusPolicyViolation Status = Status("POLICY VIOLATION") //nolint:errname // Not an error
-	StatusUnknown         Status = Status("UNKNOWN")          //nolint:errname // Not an error
+	StatusSuccess         Status = Status("SUCCESS")
+	StatusFailure         Status = Status("FAILURE")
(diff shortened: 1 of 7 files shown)


pkg/platform/reporter.go
```diff
@@ -0,0 +1,145 @@
...
+func markdownPill(text string) string {
+	return fmt.Sprintf("**`%s`**", text)
+}
+
+// markdownURL returns a markdown URL string given a title and a URL.
+func markdownURL(text, URL string) string {
+	return fmt.Sprintf("[%s](%s)", text, URL)
+}
+
+// markdonZippy returns a collapsible section with a given title and body.
```
sailorlqh, line 61:
```suggestion
// markdownZippy returns a collapsible section with a given title and body.
```

pkg/platform/reporter.go
```diff
@@ -0,0 +1,145 @@
...
+	tildeChanged = regexp.MustCompile(
+		"(?m)" + // enable multi-line mode
+			"^([\t ]*)" + // only match tilde at start of line, can lead with tabs or spaces
+			"([~])") // tilde represents changes and needs switched to exclamation for git diff
+
+	swapLeadingWhitespace = regexp.MustCompile(
+		"(?m)" + // enable multi-line mode
+			"^([\t ]*)" + // only match tilde at start of line, can lead with tabs or spaces
+			`((\-(\/\+)*)|(\+(\/\-)*)|(!))`) // match characters to swap whitespace for git diff (+, +/-, -, -/+, !)
+)
```
sailorlqh, line 81: I think moving this to top of this file would be preferred

sailorlqh (approved):

PR #444, opened 2024-12-05
Add job logs

gjonathanhong:
This is mostly duplicate code from the GitHubReporter and `pkg/github` package that will now live in the Platform interface.

### DIFF
--- a/pkg/platform/github.go
+++ b/pkg/platform/github.go
@@ -100,6 +100,17 @@ func NewGitHub(ctx context.Context, cfg *gh.Config) (*GitHub, error) {
 		graphqlClient: graphqlClient,
 	}
 
+	if cfg.GitHubServerURL != "" || cfg.GitHubRunID > 0 || cfg.GitHubRunAttempt > 0 {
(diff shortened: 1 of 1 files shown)


pkg/platform/github.go
```diff
@@ -100,6 +100,17 @@ func NewGitHub(ctx context.Context, cfg *gh.Config) (*GitHub, error) {
 		graphqlClient: graphqlClient,
 	}
 
+	if cfg.GitHubServerURL != "" || cfg.GitHubRunID > 0 || cfg.GitHubRunAttempt > 0 {
+		g.logURL = fmt.Sprintf("%s/%s/%s/actions/runs/%d/attempts/%d", cfg.GitHubServerURL, cfg.GitHubOwner, cfg.GitHubRepo, cfg.GitHubRunID, cfg.GitHubRunAttempt)
+	}
+
+	if cfg.GitHubJobName != "" {
+		resolvedURL, err := g.resolveJobLogsURL(ctx)
+		if err == nil {
```
sailorlqh, line 109: Just want to make sure we are ok with not doing anything about this err?

sailorlqh (approved):

pkg/platform/github.go
gjonathanhong: Yes, it's intentional. If there's an error getting the job URL, then it fallbacks to the URL to the entire workflow run (see the assignment directly above).

PR #450, opened 2024-12-09
Add ClearReports implementation for GitHub and Mock platform types

gjonathanhong:


pkg/platform/github.go
```diff
@@ -524,6 +525,108 @@ func (g *GitHub) CommentEntrypointsSummary(ctx context.Context, p *EntrypointsSu
 	return nil
 }
 
+// ClearComments clears any existing reports that can be removed.
+func (g *GitHub) ClearComments(ctx context.Context) error {
```
verbanicm, line 529: I would abstract this to just clear or clear status or something non-platform specific, it may not always be a "comment"

sailorlqh (approved):

PR #455, opened 2024-12-12
Add `--skip-reporting` flag to commands using reporting

gjonathanhong:


pkg/commands/apply/apply.go
```diff
@@ -123,6 +124,14 @@ func (c *ApplyCommand) Flags() *cli.FlagSet {
 		Usage:   "The duration Terraform should wait to obtain a lock when running commands that modify state.",
 	})
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "skip-reporting",
+		Target:  &c.flagSkipReporting,
+		Default: false,
+		Example: "true",
+		Usage:   "Skips reporting of the plan status on the change request.",
```
verbanicm, line 132: These usage lines should probably be specific to the command, they all look like they are for plan.

verbanicm (approved):

PR #459, opened 2024-12-13
Add `GitLab.ClearReports` implementation

gjonathanhong:


pkg/platform/platform.go
```diff
@@ -78,13 +79,14 @@ type GetPolicyDataResult struct {
 
 // Report is a comment/note on an issue or change request.
 type Report struct {
-	ID   int64
+	ID   any
```
gjonathanhong, line 82: Changed because GitLab notes have type `int` and GitHub comments have type `int64`.

verbanicm (approved):

PR #463, opened 2024-12-17
fix: use merge request internal ID for GitLab API

gjonathanhong:
There are two types of ID's for merge request, `CI_MERGE_REQUEST_ID` for instance-level and `CI_MERGE_REQUEST_IID` for project-level.  

According to GitLab's predefined variables documentation, 
> The project-level IID...is the number used in the merge request URL

### DIFF
--- a/pkg/platform/gitlab.go
+++ b/pkg/platform/gitlab.go
@@ -66,15 +66,15 @@ type gitLabConfig struct {
 	GuardianGitLabToken string
 	GitLabBaseURL       string
 
-	GitLabProjectID      int
-	GitLabMergeRequestID int
+	GitLabProjectID       int
+	GitLabMergeRequestIID int
 }
 
 type gitLabPredefinedConfig struct {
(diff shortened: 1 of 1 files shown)


pkg/platform/gitlab.go
```diff
@@ -66,15 +66,15 @@ type gitLabConfig struct {
...
 
 type gitLabPredefinedConfig struct {
-	CIJobToken       string
-	CIServerHost     string
-	CIProjectID      int
-	CIMergeRequestID int
+	CIJobToken        string
+	CIServerHost      string
+	CIProjectID       int
+	CIMergeRequestIID int
```
verbanicm, line 77: Can you add a comment similar to the PR body explaining why, doc link, etc

verbanicm (approved):

pkg/platform/gitlab.go
gjonathanhong: Added

PR #466, opened 2025-01-02
Add ListChangeRequestsByCommit skeleton to Platform interface

gjonathanhong:
This method is required for Guardian Apply to report its status on a change request. For GitHub and Guardian, the PR/MR ID will not be available from the GitHub Action/GitLab Pipeline environment when merging, so the ID needs to be looked up based on the SHA of the commit.

### DIFF
--- a/pkg/platform/gitlab.go
+++ b/pkg/platform/gitlab.go
@@ -385,3 +385,8 @@ func (g *GitLab) withRetries(ctx context.Context, retryFunc retry.RetryFunc) err
 	}
 	return nil
 }
+
+// ListChangeRequestsByCommit lists the merge requests associated with a commit SHA.
+func (g *GitLab) ListChangeRequestsByCommit(ctx context.Context, sha string) (any, error) {
+	return nil, nil
+}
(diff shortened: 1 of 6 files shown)


sailorlqh (approved):

pkg/platform/gitlab.go
```diff
@@ -385,3 +385,8 @@ func (g *GitLab) withRetries(ctx context.Context, retryFunc retry.RetryFunc) err
 	}
 	return nil
 }
+
+// ListChangeRequestsByCommit lists the merge requests associated with a commit SHA.
+func (g *GitLab) ListChangeRequestsByCommit(ctx context.Context, sha string, opts *ListChangeRequestsByCommitOptions) (any, error) {
```
capri-xiyue, line 390: Seems like previously in abc, we usually use option in such pattern https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/audit/client.go#L107C37-L107C48

capri-xiyue (approved):

pkg/platform/gitlab.go
gjonathanhong: It seems to solve a different problem - client options vs. options specific to a client method. The `ListChangeRequestsByCommitOptions` type is used to embed types defined by GitHub, GitLab, etc.

PR #468, opened 2025-01-09
Implement GitHub ListChangeRequestsByCommit

gjonathanhong:


pkg/platform/github.go
```diff
@@ -696,7 +724,50 @@ func (g *GitHub) resolveJobLogsURL(ctx context.Context) (string, error) {
...
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list pull request comments: %w", err)
+	}
+
+	return &ListChangeRequestsByCommitResponse{PullRequests: pullRequests, Pagination: pagination}, nil
+}
+
+func maybeRetryable(resp *github.Response, err error) (error, bool) {
```
verbanicm, line 768: we should centralize this somewhere over time, I believe we use this pattern all over.

verbanicm (approved):

pkg/platform/github.go
sethvargo: I mean... whoever came up with this original idea is _clearly_ a genius. But now that I see it used, you could go even further if you wanted:

```go
func maybeRetryable(resp *github.Response, err error) error {
  if _, ok := ignoredStatusCodes[resp.StatusCode]; !ok {
    return retry.RetryableError(err)
  }
  return err
}
```

then:

```go
if err != nil {
  return maybeRetryable(resp, fmt.Errorf("failed to do the thing: %w", err))
}
```

I actually think that's better, because right now you aren't adding context to the retryable error.

gjonathanhong: Thank you 😮

gjonathanhong: I hope this no longer makes you sad 😆

sethvargo: Thank you

sethvargo (approved):

PR #477, opened 2025-03-05
feat: add modifiers to allow users to force a directory to be planned/applied

verbanicm:
NOTE: This will need to be cherry picked into the v2 branch as v3 isnt released yet

### DIFF
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -231,7 +231,7 @@ func (c *EntrypointsCommand) Process(ctx context.Context) error {
 
 	metaPlanDirs, ok := metaValues[modifiers.MetaKeyGuardianDir]
 	if !ok {
-		metaPlanDirs = []string{}
+		metaPlanDirs = make([]string, 0)
 	}
 
 	for _, metaPlanDir := range metaPlanDirs {
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/abcxyz/guardian/internal/metricswrap"
 	"github.com/abcxyz/guardian/pkg/git"
+	"github.com/abcxyz/guardian/pkg/modifiers"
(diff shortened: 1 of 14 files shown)


pkg/commands/entrypoints/entrypoints.go
```diff
@@ -219,6 +220,30 @@ func (c *EntrypointsCommand) Process(ctx context.Context) error {
...
+		return fmt.Errorf("failed to find modifier content: %w", err)
+	}
+	logger.DebugContext(ctx, "modifier content", "value", modifierContent)
+
+	metaValues := modifiers.ParseBodyMetaValues(ctx, modifierContent)
+	logger.DebugContext(ctx, "parsed body meta values", "values", metaValues)
+
+	metaPlanDirs, ok := metaValues[modifiers.MetaKeyGuardianDir]
+	if !ok {
+		metaPlanDirs = []string{}
```
sethvargo, line 234:
```suggestion
		metaPlanDirs = make([]string, 0, 4)
```

pkg/modifiers/modifiers.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// metaValuesKVRegex is a regex to split key value pairs for comment modifiers.
+	metaValuesKVRegex = regexp.MustCompile(`^(GUARDIAN_[A-Z0-9_]+)=(.*)$`)
+)
+
+// MetaValues is a map of key value pairs to store meta values.
+type MetaValues map[string][]string
+
+// ParseBodyMetaValues parses the modifiders from a contents string.
+func ParseBodyMetaValues(ctx context.Context, contents string) MetaValues {
```
sethvargo, line 39: This is fine, but I do wonder how much this is going to buy us in the long run versus a simple `if strings.Contains("GUARDIAN_DIR")`. Seems like a lot of complexity based on the intention that we _might_ need more modifiers in the future?

verbanicm: Yea, but this code existed so i just pasted it back in. The effort was very minimal.

sethvargo (approved):

PR #478, opened 2025-03-06
feat: add modifiers to allow users to force a directory to be planned/applied

verbanicm:
This is cherry picked from main and updated to use code available in this branch. (platform APIs are different between main/v3 and this branch)

### DIFF
--- a/pkg/platform/github.go
+++ b/pkg/platform/github.go
@@ -403,6 +403,68 @@ func (g *GitHub) withRetries(ctx context.Context, retryFunc retry.RetryFunc) err
 	return nil
 }
 
+// ModifierContent returns the pull request body as the content to parse modifiers
+// from.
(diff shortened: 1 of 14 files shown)


pkg/platform/github.go
```diff
@@ -398,6 +398,68 @@ func (g *GitHub) withRetries(ctx context.Context, retryFunc retry.RetryFunc) err
 	return nil
 }
 
+// ModifierContent returns the pull request body as the content to parse modifiers
+// from.
+func (g *GitHub) ModifierContent(ctx context.Context) (string, error) {
```
verbanicm, line 403: This function uses the github client instead of wrapper functions within the struct like main does because they dont exist in this branch.

sethvargo (approved):
You'll need to bump the Go version to use t.Context()

PR #480, opened 2025-03-06
chore: update release process

verbanicm:
The previously release process didnt work for release/** branches.

Instead of relying on creating a tag, the new process allows users to generate a release preview in a pull request. Upon approval and push, the release will be triggered.

This process is much cleaner than the previous process and mimics what is done for jvs and pkg.

Decided to stick with goreleaser for now.


sethvargo (approved):

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,101 @@
...
+              tag_name: `v${process.env.NEXT_VERSION}`,
+              previous_tag_name: `v${process.env.CURRENT_VERSION}`,
+            });
+            core.setOutput('release-notes', releaseNotesResponse.data.body)
+
+      - name: 'Update Pull Request'
+        uses: 'abcxyz/pkg/.github/actions/create-pull-request@main' # ratchet:exclude
+        with:
+          token: '${{ steps.mint-token.outputs.token }}'
+          base_branch: '${{ github.ref_name }}'
```
verbanicm, line 96: I noticed a bug elsewhere, where this is set to default branch, this will prevent using release/** branches. This should be ref_name which equals the branch from where the workflow is triggered from for a workflow_dispatch.

shankiyani (approved):

PR #510, opened 2025-03-21
feat: add workflow subcommand for merge-check

dcreey:


pkg/commands/workflows/plan_status_comments.go
```diff
@@ -43,14 +43,14 @@ type PlanStatusCommentCommand struct {
...
 func (c *PlanStatusCommentCommand) Desc() string {
-	return `Remove previous Guardian plan comments from a pull request`
+	return `Report the status of a guardian plan`
 }
 
 func (c *PlanStatusCommentCommand) Help() string {
 	return `
 Usage: {{ COMMAND }} [options]
 
-	Remove previous Guardian plan comments from a pull request.
```
verbanicm, line 53: nit: Guardian

pkg/commands/workflows/plan_status_comments.go
```diff
@@ -43,14 +43,14 @@ type PlanStatusCommentCommand struct {
 }
 
 func (c *PlanStatusCommentCommand) Desc() string {
-	return `Remove previous Guardian plan comments from a pull request`
```
verbanicm, line 46: nit: Guardian

pkg/commands/workflows/plan_status_comments.go
dcreey: ? Isn't that what it is?

pkg/commands/workflows/merge_queue_status_comments.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (c *MergeQueueStatusCommentCommand) Flags() *cli.FlagSet {
+	set := c.NewFlagSet()
+
+	c.platformConfig.RegisterFlags(set)
+
+	f := set.NewSection("COMMAND OPTIONS")
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "failed",
```
verbanicm, line 61: Is this necessary? Seems like this command is currently only run when theres a failure.

pkg/commands/workflows/merge_queue_status_comments.go
```diff
@@ -0,0 +1,121 @@
...
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "target-branch",
+		Target:  &c.flagTargetBranch,
+		Example: "main",
+		Usage:   "The target branch that will be merged into.",
+	})
+
+	f.BoolVar(&cli.BoolVar{
```
verbanicm, line 75: Same, this command seems highly specific to queue failures, when would this be used?

pkg/commands/workflows/merge_queue_status_comments.go
dcreey: Perhaps its a yagni, but I thought I'd mirror the behavior of plan-status. In case we want to also add a comment in other scenarios (e.g. success)

pkg/commands/workflows/merge_queue_status_comments.go
dcreey: Its not, but again it seems common for this to be on every subcommand. Thought it made sense to include it.

PR #518, opened 2025-04-03
chore: bump guardian and change default log env vars

dcreey:


cmd/guardian/main.go
```diff
@@ -174,15 +175,19 @@ func realMain(ctx context.Context) error {
 // setLogEnvVars set the logging environment variables to their default
 // values if not provided.
 func setLogEnvVars() {
-	if os.Getenv("GUARDIAN_LOG_FORMAT") == "" {
-		os.Setenv("GUARDIAN_LOG_FORMAT", defaultLogFormat)
+	if os.Getenv("LOG_FORMAT") == "" {
```
verbanicm, line 178: should we be checking both sets of env vars instead of the global ones?

verbanicm: or is there a better pattern here?

dcreey: No, because the custom GUARDIAN_ one will still take precedent. This is the fallback if the user doesn't provide those.

verbanicm: ah got ya

sailorlqh (approved):

verbanicm (approved):

verbanicm (changes requested):
wait for decision in https://github.com/abcxyz/tagrep/pull/22#discussion_r2027620545

verbanicm (approved):

PR #522, opened 2025-04-11
chore: remove lint workflows

dcreey:
Updating so that we use the lint.yml workflow from pkg https://github.com/abcxyz/pkg/blob/main/.github/workflows/lint.yml. This workflow is required via an org ruleset so that it runs on this repo.


sethvargo (approved):

verbanicm (approved):
@dcreey Somethings up with the way the checks are being done...
I removed the old required ones

PR #525, opened 2025-04-17
Allow configuring the base URL of GitHub Enterprises

gjonathanhong:
Currently, Guardian is configured to only call the GitHub SaaS API so it can only be run on GitHub Actions within the GitHub.com. This change exposes some options to modify the GitHub base URLs so that it can support GitHub Enterprise instances. Most of these will be inferred from the GitHub Actions environment from the default environment variables set by GitHub.

### DIFF
--- a/pkg/github/config.go
+++ b/pkg/github/config.go
@@ -34,6 +34,7 @@ type Config struct {
 	// Enterprise
 	GitHubAPIURL     string
 	GitHubGraphQLURL string
+	GitHubServerURL  string
 
 	// Auth
 	GuardianGitHubToken     string
@@ -46,7 +47,6 @@ type Config struct {
 	Permissions             map[string]string
 
 	GitHubEventName         string
-	GitHubServerURL         string
 	GitHubRunID             int64
 	GitHubRunAttempt        int64
 	GitHubJob               string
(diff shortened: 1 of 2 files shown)


pkg/github/config.go
```diff
@@ -31,6 +31,10 @@ type Config struct {
 	InitialRetryDelay time.Duration
 	MaxRetryDelay     time.Duration
 
+	// Enterprise
```
sethvargo, line 34: I think you're also going to want the HTTP (web) URL for links

gjonathanhong: I think that exists already as `GitHubServerURL`. Moved it up to the same grouping

pkg/github/config.go
```diff
@@ -31,6 +31,11 @@ type Config struct {
...
+	GitHubAPIURL     string
+	GitHubGraphQLURL string
+	GitHubServerURL  string
```
verbanicm, line 37: Should we default this too or is that done somewhere in some other library?

pkg/github/config.go
sethvargo: Ah thanks!

sethvargo (approved):

pkg/github/config.go
gjonathanhong: I just set the default to `https://github.com`. Most of these will be inferred from the default environment variables provided by the GitHub Actions.

sethvargo (approved):

PR #526, opened 2025-04-17
chore: add ratchet and action linters

dcreey:

PR #532, opened 2025-04-18
fix broken release workflow

gjonathanhong:
This change proposes a fix to address the [broken release workflow](https://github.com/abcxyz/guardian/actions/runs/14525816461/job/40757050540#step:5:34) due to recent [linting updates](https://github.com/abcxyz/guardian/pull/526).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -191,10 +191,10 @@ jobs:
 
           while IFS= read -r -d '' BINARY; do
             chmod +x "${BINARY}"
-            echo $BINARY
+
             FOLDER="$(basename "${BINARY}" ".exe")"
             NAME="${BINARY%%_*}"
-            echo $NAME
+
             # Windows - zip and follow Windows conventions with named subdirectory
             if [ "${FOLDER}" != "${BINARY}" ]; then
               NAME="${NAME}.exe"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -213,8 +213,8 @@ jobs:
           wait
 
           FILES="$(find ./* -name '*.tar.gz' -o -name '*.zip')"
-          sha256sum "${FILES}" > SHA256SUMS
-          sha512sum "${FILES}" > SHA512SUMS
+          sha256sum ''${FILES}'' > SHA256SUMS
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -209,12 +209,11 @@ jobs:
               mv "${BINARY}" "${NAME}"
               (tar -czvf "${FOLDER}.tar.gz" "${NAME}" && rm -rf "${NAME}") &
             fi
-          done <   <(find ./* -type f)
+          done <   <(find . -type f -printf '%P\0')
```
gjonathanhong, line 212: The `while` loop expects a NULL delimiter (represented by `\0`), but the default behavior of `find` is to delimit using `\n`. This means that previously it would treat filenames delimited by newlines as a single argument, so it would only evaluate to a single iteration of the loop. To fix this, I added a `-printf` option to align on the NULL delimiter.


Additionally, we added `./` to the search pattern in the `find` command and this changed the behavior to include the relative path in the result. This messed with evaluation of conditions within the while loop block, so I also used `%P` in the `printf` option, to only print the filename without the relative path.

.github/workflows/release.yml
```diff
@@ -209,12 +209,11 @@ jobs:
...
           wait
 
-          FILES="$(find ./* -name '*.tar.gz' -o -name '*.zip')"
-          sha256sum "${FILES}" > SHA256SUMS
-          sha512sum "${FILES}" > SHA512SUMS
+          find . \( -name '*.tar.gz' -o -name '*.zip' \) -type f -printf '%P\0' | xargs -0 sha256sum > SHA256SUMS
```
gjonathanhong, line 215: Due to recent linting rules, `$()` and `${}` expressions are wrapped in double quotation marks. This changed the format/behavior of  `FILES` and the filenames were passed into `sha256sum and `sha512sum` as a single argument, filenames delimited by newline characters. 

This doesn't work because `sha256sum and `sha512sum` expect filenames to be in separate arguments, so I'm using `xargs` to pass the result of `find` directly to the checksum commands. This is the only solution I found to satisfy the linting requirements (and also the linter doesn't provide an easy way to ignore checks), but open to other suggestions.

sethvargo (approved):

PR #535, opened 2025-04-22
chore: update default arrayification for guardian admin

dcreey:
Parse input as array in order to correctly escape.


sethvargo (approved):
"arrayification"

PR #545, opened 2025-06-11
Use reusable guardian plan workflow

RivocPX:
Replace `guardian-plan` workflow with `call-guardian-plan` and `reusable-guardian-plan`.

Previous guardian-plan workflow is triggered by `pull_request_target`, which grants the workflow higher permission. And it’s considered bad practice and can lead to security issues when executing untrusted codes. It will also raise code scanning alerts.
This PR makes `guardian-plan` workflow reusable(`reusable-guardian-plan`). And it’s called by a caller workflow `call-guardian-plan`. `call-guardian-plan` is triggered by `pull_request`, and it will only use the `reusable-guardian-plan` on the main branch. Thus, security issues could be prevented.
Workflows are tested [here](https://github.com/abcxyz/guardian-staging/pull/90). (See workflows in`All checks have passed`)
Note that migrate from current `guardian-plan` to reusable one may require additional settings like changing attribute condition and minty.


pdewilde:
Forgive my ignorance, but why do we need two workflows, one to call the other?

ian-shafer:
> Forgive my ignorance, but why do we need two workflows, one to call the other?

@pdewilde We need this because of the use of `pull_request_target` and code scanning. See here: https://github.com/theteamatx/x-refinery-org/security/code-scanning/1.

@RivocPX Can you add this to the PR description. This is context that should be included there.

ian-shafer:
> @RivocPX Can you add this to the PR description. This is context that should be included there.

Sorry, you've already included this. Please ignore.

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
```diff
@@ -181,7 +172,7 @@ jobs:
 
   plan_success:
     if: |
-      always() && github.event_name == 'pull_request_target'
+      always()
```
ian-shafer, line 175: Looks like this will always (pun intended) succeed, so we can just remove the `if` statement.

abc.templates/base-workflows/testdata/golden/basic/data/.github/workflows/reusable-guardian-plan.yml
```diff
@@ -181,7 +172,7 @@ jobs:
 
   plan_success:
     if: |
-      always() && github.event_name == 'pull_request_target'
+      always()
```
ian-shafer, line 175: Same as above. We can get rid of this `if`.

abc.templates/base-workflows/contents/workflows/call-guardian-plan.yml
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Call Guardian Plan is used to call Reusable Guardian Plan workflow
+name: 'call_guardian_plan'
+on:
+  pull_request:
```
ian-shafer, line 18: Should this be `pull_request_target`?

ian-shafer:
> That doesn't quite answer my question. We already changed from pull_request_target to pull_request, so why do we need two separate files?

I've got an open question about that. I think we have to keep it as `pull_request_target`.

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
verbanicm: i think wee need to keep this but make it pull_request, this was added to support the mrege_group event i believe

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
```diff
@@ -121,7 +112,7 @@ jobs:
 
   plan:
     if: |
-      github.event_name == 'pull_request_target' && needs.init.outputs.directories != '[]'
+      needs.init.outputs.directories != '[]'
```
verbanicm, line 115: i think wee need to keep this but make it pull_request, this was added to support the mrege_group event i believe

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
verbanicm, line 1: nit: not sure i like this name, maybe just keep it guardian-plan.yml

abc.templates/base-workflows/contents/workflows/call-guardian-plan.yml
ian-shafer: FYI, I could be completely wrong here... I'm trying to remember how we wanted this to work.

verbanicm: i dont think, the goal was to get rid of pull_request_target. we need the plan run to come from main, not necessarily this file

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
```diff
@@ -47,8 +40,6 @@ jobs:
     # We do not run any plan jobs in the merge queue. But these jobs are still
     # required to run (in a skipped state) in order to block adding to the
     # merge queue until a plan has been created.
-    if: |
```
verbanicm, line 50: i think wee need to keep this but make it pull_request, this was added to support the mrege_group event i believe

abc.templates/base-workflows/contents/workflows/reusable-guardian-plan.yml
RivocPX: @ian-shafer would you like me to change it from reusable-guardian-plan to guardian-plan?

abc.templates/base-workflows/contents/workflows/call-guardian-plan.yml
```diff
@@ -0,0 +1,37 @@
...
+  merge_group:
+
+jobs:
+  call_guardian_plan:
+    permissions:
+      contents: 'read'
+      pull-requests: 'write'
+      id-token: 'write'
+    name: 'Call reusable guardian plan'
+    uses: 'REPLACE_GITHUB_REPOSITORY/.github/workflows/reusable-guardian-plan.yml@main' # ratchet:exclude
```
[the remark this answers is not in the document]
RivocPX: I think this won't work. We have WIF authentication and attribute condition like [this](https://github.com/abcxyz/infra-gcp/blob/main/terraform/workload-identity/github/abcxyz/guardian-staging/guardian_guardian-staging.tf#L73-L74) will prevent this situation.

abc.templates/base-workflows/contents/workflows/call-guardian-plan.yml
ian-shafer: I was wrong. `pull_request` is correct here. `pull_request` is inherently safer because it does has read-only access to the repo and no access to secrets (as opposed to `pull_request_target`). So we want to use `pull_request` whenever possible.

ian-shafer:
> That doesn't quite answer my question. We already changed from pull_request_target to pull_request, so why do we need two separate files?

pull_request does not give write access to the PR. This is why pull_request_target was used in the first place (for write access to the PR). But, the guardian plan workflow also requires a checkout of the head branch (untrusted code). This is why the code scanner alerted on it. So we've moved the code that needs to write to the PR to a separate workflow file that will always execute from the main branch.

RivocPX:
> > > Forgive my ignorance, but why do we need two workflows, one to call the other?
> > 
> > 
> > @pdewilde We need this because of the use of `pull_request_target` and code scanning
> 
> That doesn't quite answer my question. We already changed from pull_request_target to pull_request, so why do we need two separate files?

I think we need two separate files because we ensure only guardian-plan on main branch will be executed. If in one file, there is possibility to checkout to malicious codes.

sethvargo:
I'm not sure what we're trying to achieve with this PR anymore, but it seems to be adding significant complexity.

ian-shafer:
> I'm not sure what we're trying to achieve with this PR anymore, but it seems to be adding significant complexity.

Yes, you may be correct. The goal here was to avoid alerts from CodeQL scanning with pull_request_target mixed with a checkout of the head (untrusted) code.

To be clear, the alert that we see from CodeQL is calling out a _potential_ issues, but it is not an _actual_ security issue in Guardian. So the changes here were to simply re-organize the code to avoid the alert.

But this has resulted in the added complexity.

We're now leaning toward documenting this as a known issue and potentially giving guidance on how to ignore this in CodeQL scans.

RivocPX:
We will write a follow-up document about this. This PR will be closed for now.

PR #576, opened 2026-03-06
attempt to find the modules.json file while checking providers

crwilcox:
also walk those submodule directories for their provider usage

### DIFF
--- a/pkg/checkterraform/checkterraform.go
+++ b/pkg/checkterraform/checkterraform.go
@@ -17,9 +17,11 @@ package checkterraform
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/fs"
 	"maps"
+	"os"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -40,6 +42,15 @@ type ScanResult struct {
 	Provisioners        []string
 	InvalidProviders    []string
 	InvalidProvisioners []string
+}
+
+// moduleJSON is the structure of the .terraform/modules/modules.json file.
+type moduleJSON struct {
+	Modules []struct {
+		Key    string `json:"Key"`
+		Source string `json:"Source"`
+		Dir    string `json:"Dir"`
+	} `json:"Modules"`
 }
 
 // findInvalid returns the list of items in actual that are not in allowed, or if allowed is empty,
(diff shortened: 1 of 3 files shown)


crwilcox:
While this doesn't have much in the way of testing atm, I did some inspecting of structure.

If we have an instance, using a module (github_org), which then also uses a module (github_repo), here is the output of modules.json 

```
❯ cat .terraform/modules/modules.json
{"Modules":[{"Key":"","Source":"","Dir":"."},{"Key":"org_resources","Source":"../../modules/github-org","Dir":"../../modules/github-org"},{"Key":"org_resources.nested_repo_settings","Source":"../../modules/github-repo","Dir":"../../modules/github-repo"}]}%
```

bradegler (approved):

pkg/checkterraform/checkterraform.go
```diff
@@ -223,3 +238,34 @@ func extractRequiredProviders(body hcl.Body, providerSet map[string]struct{}) {
...
 	}
 }
+
+// extractPathsFromModulesJSON extracts the paths to walk from modules.json if present.
+func extractPathsFromModulesJSON(ctx context.Context, dir string) []string {
+	logger := logging.FromContext(ctx)
+	pathsToWalk := []string{dir}
+
+	modulesJSONPath := filepath.Join(dir, ".terraform", "modules", "modules.json")
+	if content, err := os.ReadFile(modulesJSONPath); err == nil {
```
jalseth: Please reduce nesting.

```go
content, err := os.ReadFile
if err != nil {
 warnlog
  return pathsToWalk
}

if err := json.Marshal; err != nil {
  warnlog
  return pathsToWalk
}

for _, m ... {
}

return pathsToWalk
```

pkg/terraform/plan.go
```diff
@@ -90,5 +90,50 @@ func (t *TerraformClient) Plan(ctx context.Context, stdout, stderr io.Writer, op
 		return 0, fmt.Errorf("encountered an error while checking terraform: %w", err)
 	}
 
+	// check for modules.json, created by init.
```
jalseth: Is this relevant? The test code was not updated here.

jalseth (approved):

pkg/terraform/plan.go
crwilcox: lol. no. this was me noodling when I started the patch *facepalm*

pkg/checkterraform/checkterraform.go
crwilcox: Good call :)

PR #581, opened 2026-04-07
fix: Remove empty strings from allow/deny list flags

jalseth:
Calling `strings.Split()` on an empty string returns `[]string{""}`, not `[]string{}` so when this flag was set to an empty list it blocked all usage of Guardian.

### DIFF
--- a/pkg/checkterraform/checkterraform.go
+++ b/pkg/checkterraform/checkterraform.go
@@ -66,6 +66,12 @@ func findInvalid(allowed, disallowed, actual []string) []string {
 	return slices.DeleteFunc(result, func(s string) bool { return !slices.Contains(disallowed, s) })
 }
 
+func removeBlankStrings(s []string) []string {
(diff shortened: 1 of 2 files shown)


pkg/checkterraform/checkterraform.go
```diff
@@ -66,6 +66,12 @@ func findInvalid(allowed, disallowed, actual []string) []string {
 	return slices.DeleteFunc(result, func(s string) bool { return !slices.Contains(disallowed, s) })
 }
 
+func removeEmptyStrings(s []string) []string {
```
lock14: small nit: Consider renaming this to `removeBlankStrings`? Since it uses TrimSpace, it removes whitespace-only strings in addition to truly empty ones, and "blank" typically implies both.

lock14 (approved):

crwilcox (approved):

pkg/checkterraform/checkterraform.go
jalseth: Done.

PR #595, opened 2026-06-11
fix(entrypoints): Walk up directory path for all changed files looking for entry points

### LINKED ISSUE #593: entrypoints: command with -detect-changes fails if Terraform files are not being modified
### TL;DR

If you only change non-Terraform files, Guardian does not trigger a plan even when there are Terraform configs which depend on those via `fileset()`.

### Expected behavior

Guardian detects the file change and triggers a plan in the PR.

### Observed behavior

Guardian outputs an empty list `[]` of changed files which causes the plan to be skipped.

### Environment Details

```markdown
Guardian Version: 3.1.1
```

### Additional information

*The info below was produced by Gemini*

## Steps to Reproduce
1. Create an entrypoint with Terraform configuration that references files in a subdirectory using `fileset()` (e.g., `fileset(path.module, "configs/*.yaml")`).
2. Modify one of these configuration files (e.g., `my-entrypoint/configs/app.yaml`).
3. Run `guardian entrypoints -detect-changes` between commits.
4. **Result**: It returns `[]` instead of `["my-entrypoint"]`.

## Root Cause
In `pkg/commands/entrypoints/entrypoints.go` -> `detectEntrypointChanges`:
It performs exact matching on the directories returned by `git diff` against the keys of `EntrypointToModules` and `ModulesToEntrypoints`. A changed subdirectory path (e.g., `my-entrypoint/configs`) will not match the entrypoint root path (`my-entrypoint`).

## Suggested Fix
Modify `detectEntrypointChanges` to walk up the directory tree from each changed path to check if any parent directory is a known entrypoint or module.

jalseth:
This resolves the issue where files included via fileset() under an entry point's directory were ignored unless Terraform files were also being changed.

---

Fixes https://github.com/abcxyz/guardian/issues/593

### DIFF
--- a/pkg/commands/entrypoints/entrypoints.go
+++ b/pkg/commands/entrypoints/entrypoints.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io/fs"
+	"path/filepath"
 	"slices"
 
 	"golang.org/x/exp/maps"
@@ -49,6 +50,7 @@ type EntrypointsCommand struct {
 	flagDetectChanges           bool
 	flagFailUnresolvableModules bool
 	flagMaxDepth                int
+	flagMaxWalkbackDepth        int
 	flagSkipReporting           bool
 
 	parsedFlagMaxDepth *int
@@ -57,6 +59,11 @@ type EntrypointsCommand struct {
 
 	newGitClient func(ctx context.Context, dir string) git.Git
 }
+
+const (
+	// Set a maximum limit to ensure we have an upper bound on how much time we spend.
+	maxWalkbackDepthLimit = 100
+)
 
 func (c *EntrypointsCommand) Desc() string {
 	return `Determine the entrypoint directories to run Guardian commands`
@@ -122,6 +129,13 @@ func (c *EntrypointsCommand) Flags() *cli.FlagSet {
 		Target:  &c.flagMaxDepth,
 		Usage:   `How far to traverse the filesystem beneath the target directory for entrypoints.`,
 		Default: -1,
+	})
+
+	f.IntVar(&cli.IntVar{
+		Name:    "max-walkback-depth",
+		Target:  &c.flagMaxWalkbackDepth,
+		Usage:   fmt.Sprintf("How far to walk up the directory tree to find a parent entrypoint or module for changed files. Cannot exceed %d. Zero disables this feature.", maxWalkbackDepthLimit),
+		Default: 5,
 	})
 
 	f.BoolVar(&cli.BoolVar{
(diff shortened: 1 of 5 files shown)


shankiyani (approved):

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -301,14 +337,18 @@ func (c *EntrypointsCommand) detectEntrypointChanges(ctx context.Context, dir st
...
-			for entrypoint := range entrypoints {
-				modifiedEntrypoints[entrypoint] = struct{}{}
-			}
+	for _, d := range diffDirs {
+		if !strings.HasPrefix(d, dir) {
+			continue
+		}
+
+		matchedDir, entrypoints, ok := matchClosestDirectory(d, moduleUsageGraph, c.flagMaxWalkbackDepth)
+		if !ok {
```
dcreey, line 344: This deserves a log warning.

pkg/commands/entrypoints/entrypoints.go
```diff
@@ -281,6 +296,27 @@ func (c *EntrypointsCommand) findEntrypointDirs(ctx context.Context, dir string)
 	return entrypointDirs, nil
 }
 
+// matchClosestDirectory walks up the directory tree from changedDir to find the closest matching module or entrypoint in the usage graph, up to maxWalkbackDepth.
+func matchClosestDirectory(dir string, moduleUsageGraph *terraform.ModuleUsageGraph, maxWalkbackDepth int) (string, []string, bool) {
```
dcreey: matchClosestAncestorDirectory?

pkg/commands/entrypoints/entrypoints_test.go
```diff
@@ -160,11 +161,44 @@ func TestEntrypointsProcess(t *testing.T) {
 			expStdout: `["testdata/entrypoint1/project1","testdata/entrypoint1/project2"]`,
 		},
 		{
-			name:              "no_changes_without_add_entrypoint",
+			name:              "skips_changes_outside_directory",
```
dcreey: Do we have a test for when one directory has an entrypoint and a subdirectory also has an entrypoint? I'm not sure that is a good structure but we should confirm we only get the child directory if it changes (not also the parent).

dcreey (approved):

pkg/commands/entrypoints/entrypoints.go
jalseth: Done.

pkg/commands/entrypoints/entrypoints.go
jalseth: I think it makes more sense in the matchClosestAncestorDirectory func when the max depth has been exceeded, so I've added it there.

pkg/commands/entrypoints/entrypoints_test.go
jalseth: Added.

PR #606, opened 2026-08-25
Add registry proxy support for routing Terraform providers through a network mirror

Valentin-Metz:
## Summary

Adds support for routing Terraform provider downloads through a Provider
Registry Proxy / Network Mirror via a new `--registry-proxy` flag.

This is useful in locked-down or air-gapped CI environments where direct
access to the public Terraform registry is unavailable and providers must be
fetched through an internal network mirror.

## What's included

**1. `pkg/terraform/registryproxy` package**

- `GenerateCLIConfig` produces a temporary Terraform CLI configuration
  (`.terraformrc`) containing a `provider_installation` `network_mirror`
  block that routes all providers (`"*/*/*"`) through the given proxy URL.
- Returns the config path, an always-safe cleanup function, and an error.
- When `TF_CLI_CONFIG_FILE` is already set, its contents are preserved by
  prepending them, so top-level directives (e.g. `plugin_cache_dir`,
  `credentials`) survive.
- Because Terraform permits only one `provider_installation` block, an
  existing config that already declares one fails fast with an actionable
  error (block detection uses `hcl/v2`).
- Proxy URLs must be `https` (Terraform only accepts https network mirror
  URLs); non-https URLs are rejected up front. An unparseable existing
  `TF_CLI_CONFIG_FILE` surfaces a parse error rather than being misreported
  as a conflicting `provider_installation` block.

**2. `--registry-proxy` flag wired into `plan`, `apply`, and `run`**

- New `RegistryProxyFlags` section exposing `--registry-proxy` (with a
  `GUARDIAN_REGISTRY_PROXY` env var fallback).
- When set, the command generates a temporary terraformrc via
  `registryproxy.GenerateCLIConfig` and passes it to Terraform through
  `TF_CLI_CONFIG_FILE`, removing the file after the command completes.
- Documented under a new Registry Proxy Options section in `cli.md`.

## Testing

- `go build ./...`
- `go test ./pkg/terraform/registryproxy/`

### DIFF
--- a/pkg/terraform/registryproxy/registryproxy.go
+++ b/pkg/terraform/registryproxy/registryproxy.go
@@ -1,0 +1,146 @@
+// Copyright 2026 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package registryproxy generates Terraform CLI configuration for Guardian, such
+// as a temporary .terraformrc that routes provider requests through a registry proxy.
+package registryproxy
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclparse"
+
+	"github.com/abcxyz/pkg/logging"
+)
+
+// GenerateCLIConfig creates a temporary Terraform CLI configuration file (.terraformrc)
+// configured to route all provider requests through the Provider Network Mirror at proxyURL.
+// The proxy is fully transparent: it decides where to fetch each provider from, so all
+// providers ("*/*/*") are routed through it. proxyURL must use the https:// scheme,
+// because Terraform only accepts https network mirror URLs.
+//
+// If TF_CLI_CONFIG_FILE is already set, its contents are preserved by prepending them to
+// the generated config, so top-level directives (such as plugin_cache_dir or credentials)
+// are not lost. Because Terraform permits only one provider_installation block, an existing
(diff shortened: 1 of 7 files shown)


pkg/terraform/registryproxy/registryproxy.go
```diff
@@ -0,0 +1,132 @@
...
+//
+// It returns the absolute file path, a cleanup function to remove the temporary file, and an error if any.
+// An empty proxyURL returns an empty path and a no-op cleanup with a nil error. The returned cleanup is
+// always non-nil and safe to call, even when an error is returned.
+func GenerateCLIConfig(ctx context.Context, proxyURL string) (string, func(), error) {
+	if proxyURL == "" {
+		return "", func() {}, nil
+	}
+
+	if !strings.HasPrefix(proxyURL, "https://") {
```
gjonathanhong:
```
u, err := url.Parse(proxyURL)
if err != nil {
    return "", func() {}, fmt.Errorf("invalid registry proxy URL: %w", err)
}
if u.Scheme != "https" {
    return "", func() {}, fmt.Errorf("registry proxy URL must use https://")
}
```

pkg/terraform/registryproxy/registryproxy.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	if !strings.HasSuffix(proxyURL, "/") {
+		proxyURL += "/"
+	}
+
+	var buf bytes.Buffer
+
+	if existing := os.Getenv("TF_CLI_CONFIG_FILE"); existing != "" {
+		data, err := os.ReadFile(existing)
```
gjonathanhong, line 76: Guardian has a `-dir` flag that sets the Terraform subprocess's working directory to a subdirectory. Do we need to ensure that this file exists at the Terraform working dir?

```
if existing := os.Getenv("TF_CLI_CONFIG_FILE"); existing != "" {
    // If the path is relative, resolve it against the workingDir
    if !filepath.IsAbs(existing) {
        existing = filepath.Join(workingDir, existing)
    }
    data, err := os.ReadFile(existing)
    // ... rest of logic
```

gjonathanhong (commented):
Please fix failed CI

pkg/terraform/registryproxy/registryproxy.go
Valentin-Metz: The file generated by us is absolute, but the the preexisting `.terraformrc` would have been resolved against Guardians workdir. Fixed and test added.

Valentin-Metz:
Closed in favor of #607, which originates from an internal branch, so integration tests and CI can authenticate via WIF and run.

